PR #6, opened 2025-09-29
Surface project

birajstha:
Related to 
- [] https://github.com/users/tfunck/projects/1/views/1?pane=issue&itemId=115217456&issue=tfunck%7Cneuromaps-nhp-prep%7C8
- [] https://github.com/users/tfunck/projects/1/views/1?pane=issue&itemId=119556933&issue=tfunck%7Cneuromaps-nhp-prep%7C13
- [] https://github.com/users/tfunck/projects/1/views/1?pane=issue&itemId=98571209&issue=tfunck%7Cneuromaps-nhp-prep%7C4
- [] https://github.com/users/tfunck/projects/1/views/1?pane=issue&itemId=120333391&issue=tfunck%7Cneuromaps-nhp-prep%7C14

[Proposed Architecture](https://excalidraw.com/#json=lgiIv_4ESyTHA2Fjk2a_L,AQQ-alLIWB8kb242xtrnYA) 
image


src/neuromaps_nhp/graph/transforms.py
```diff
@@ -0,0 +1,187 @@
...
+        # direct transform
+        filepath = resource_manager.get_filepath(
+            resource_type="transform",
+            source=source,
+            target=target,
+            density=density,
+            hemisphere=hemisphere,
+            resource_name=resource_name
+        )
+        if filepath is None:
```
kaitj, line 23: Ahh I see, I missed this in the other module.

src/neuromaps_nhp/graph/transforms.py
```diff
@@ -0,0 +1,187 @@
...
+    show_graph=False
+    
+    specific_results = search_transforms(
+        source=source,
+        target=target,
+        density=density,
+        hemisphere=hemisphere,
+        resource_name="sphere",
+    )
+    print(f"Found {len(specific_results['available_transforms'])} transforms for source {source}, target {target}, {density} density, {hemisphere} hemisphere")
```
kaitj, line 187: This should be part of a test

src/neuromaps_nhp/graph/transforms.py
```diff
@@ -0,0 +1,187 @@
...
+        print(f"  {path_name} (steps: {path_info['length']})")
+        print(f"    Route: {path_str}")
+    
+    print("\nAvailable Transform Files:")
+    for i, transform in enumerate(summary['available_transforms'], 1):
+        print(f"  {i}. {transform.source} -> {transform.target}")
+        print(f"     Density: {transform.density}, "
+              f"Hemisphere: {transform.hemisphere}, "
+              f"Resource: {transform.resource_name}")
+
```
kaitj, line 171: This seems like it would be useful for debugging. I wonder if it makes sense to break this down into smaller pieces that can be called individually?

src/neuromaps_nhp/graph/transforms.py
kaitj, line 1: Could you walk me through the module here? There are some defined methods here that are already defined as in another module.

tests/test_fetch_resource.py
```diff
@@ -0,0 +1,87 @@
...
+def test_fetch_atlas():
+    # Test fetching a known atlas resource
+    atlas_path = fetch_atlas(
+        resource_name="sphere",
+        source="Yerkes19",
+        density="32k",
+        hemisphere="left"
+    )
+    assert atlas_path is not None
+    assert os.path.exists(atlas_path)
```
kaitj, line 18: Looking at this test, and then going back to the the methods, we should just return a `Path` instead of a str. From there, can just call something like `atlas_path.exists()`

Further, for this test, I would probably just test that the path exists, which would imply that the path is not None

tests/test_fetch_resource.py
```diff
@@ -0,0 +1,87 @@
+
+import pytest
+from pathlib import Path
```
kaitj, line 3:
```suggestion
```

Remove unused imports (currently)

tests/test_surface_sphere_project_unproject.py
```diff
@@ -0,0 +1,46 @@
...
+    project_to_sphere       = Yerkes19 (Intermediate)
+    unproject_from_sphere   = Yerkes19_to_D99 (Target)
+    out_sphere              = str(Path(f"{data_dir}/out_sphere.surf.gii").resolve())
+    ------------------------
+    """
+
+    result = surface_sphere_project_unproject(
+        sphere_in, sphere_project_to, sphere_unproject_from, sphere_out
+    )
+    assert os.path.exists(result)
```
kaitj, line 43: Similar to the other test, we should just return a path here and use the `path.exist()` method

tests/test_surface_sphere_project_unproject.py
```diff
@@ -0,0 +1,46 @@
...
+    
+    # check if the output file has the same number of vertices as the input file
+    assert nib.load(result).darrays[0].data.shape == nib.load(sphere_in).darrays[0].data.shape
```
kaitj, line 46: I think I commented on this in another thread and maybe I am just misremembering how this works (maybe leave a comment in the code if so) - do we expect the input arrays to be the same shape / have the same number of vertices as the output?

tests/test_surface_sphere_project_unproject.py
```diff
@@ -0,0 +1,46 @@
...
+
+from styxdefs import set_global_runner
+from styxsingularity import SingularityRunner
+from niwrap import workbench as wb
+import nibabel as nib
+
+from neuromaps_nhp.utils.surface_sphere_project_unproject import surface_sphere_project_unproject
+from neuromaps_nhp.config import config
+
+@pytest.mark.parametrize(
```
kaitj, line 13: This isn't quite how we want to use parametrize unless you are adding additional cases so that we can test multiple cases.

tests/test_surface_sphere_project_unproject.py
```diff
@@ -0,0 +1,46 @@
+import pytest
+import os
+from pathlib import Path
+
+from styxdefs import set_global_runner
+from styxsingularity import SingularityRunner
+from niwrap import workbench as wb
```
kaitj, line 7:
```suggestion
```

Unused imports

kaitj (changes requested):
I think this is a good start, left a notes:
 
- The `brain_atlas_graph.png` is a nice visualization for what we are ultimately trying to accomplish. Not sure if it makes sense to have it where it currently lives, but something to keep in mind as we may want to move it to a different folder.
- Type hints should be updated, ideally to follow the py310+  syntax.
- Before a merge, should also be run through the linter / formatter (pre-commit can be used for this too).
- There might be some redundancy with built-in `networkx` methods, so we should take a look into that as well.
- There is also a lot of calling of methods within other methods (e.g. `fetch_resources`, might call `fetch_transforms`, etc.). This is fine, as long as it makes sense / there is solid reasoning behind it.

Would be good to have a larger discussion around this at the next meeting. Another minor comment, more for the PR in general:

- Given how large the PR is, it might make sense to make it 1 commit for each feature / change for future PRs so that it is easier to follow along until we slow down on the implementation / changes. Or if it makes sense, can also do a bunch of tiny PRs to this branch

Issue #8, opened 2025-10-01
Bug: Fix pyproject.toml


kaitj:
### Description

Just noticed a mistake with the `ruff.format` options that is causing VSCode to scream.

Update `[too.ruff.format]` to `[tool.ruff.format]`

### Steps to Reproduce

Open pyproject.toml

### Expected Results

No errors

### Actual Results

Errors

PR #9, opened 2025-10-09
NiWrap wrapper for Surface sphere project unproject workbench command and related tests

birajstha:
Since [this PR #6 ](https://github.com/childmindresearch/neuromaps-nhp/pull/6) got big. I am breaking it down into multiple smaller ones.

This PR adds 
- [x] the wrapper function mentioned in [this issue #8](https://github.com/tfunck/neuromaps-nhp-prep/issues/8)
- [x] related tests

The tests still use the local paths so it fails on the CI. However, I have confirmed its passing on my local runs
image


tests/surface/test_surface_utils.py
kaitj: Were the `pytest.skip`s added for GH? If so, I would remove them and let the test fail. The files, in theory, would be available and if it needs to be skipped for other reasons, we can mark them as a skip later on.

kaitj (commented):
Looks great! Thanks for also restructuring and moving the `volumes` sub-module. Do you think `surface_utils.py` will have more utilities? Otherwise, I think we can rename then it to `utils.py` (e.g. `transforms.utils`) and have it as a utilities specific to transforms (there's not much if any for volumes).

PR #12, opened 2025-10-20
Graph

birajstha:
Related to 
- [x] [#13 Implement a data structure to represent the transformation graph](https://github.com/tfunck/neuromaps-nhp-prep/issues/13)
- [x] [#4 Create a matrix with the transformation files as elements and rows/cols as coordinate spaces](https://github.com/tfunck/neuromaps-nhp-prep/issues/4)  
for surface resources and transformations

- Implemented networkx graph and added some utility functions.

Added 
- Pydantic models for validation.
- Plotting module for graph visualization
- related tests

image

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,9 +35,7 @@ dev = [
   "ruff>=0.13.1",
   "types-pyyaml>=6.0.12.20250915"
 ]
-docs = [
-  "pdoc>=15.0.4"
-]
+docs = ["pdoc>=15.0.4"]
 
 [build-system]
 requires = ["hatchling", "hatch-vcs"]
@@ -66,8 +64,12 @@ src = ["src"]
 target-version = "py312"
 
 [tool.ruff.lint]
+ignore = [
+  "ANN002",  # ignore type hinting *args
+  "ANN003",  # ignore type hinting **kwargs
+  "D417"  # ignore docstrings for *args and **kwargs
+]
 select = ["ANN", "D", "E", "F", "I"]
-ignore = []
 fixable = ["ALL"]
 unfixable = []
 
@@ -75,7 +77,7 @@ unfixable = []
 convention = "google"
 
 [tool.ruff.lint.per-file-ignores]
-"tests/**/*.py" = []
+"tests/**/*.py" = ["ANN201", "ANN401"]
 
 [tool.ruff.format]
 quote-style = "double"
--- a/src/neuromaps_prime/utils.py
+++ b/src/neuromaps_prime/utils.py
@@ -1,0 +1,40 @@
+"""Utility functions."""
+
+from niwrap import use_docker, use_local, use_singularity
+
+
+def set_runner(
+    runner: str,
+    image_map: dict[str, str] | None = None,
+    **kwargs,
+) -> None:
+    """Set StyxRunner to use for NiWrap.
+
+    Args:
+        runner: Styx runner type to use (one of 'local', 'docker', 'singularity).
+        image_map: Optional dictionary of container paths to override.
+
+    Raises:
+        TypeError: if image_map is not dictionary.
+        ValueError: if singularity runner is used and image_map not provided.
+        NotImplementedError: if provided runner not a valid StyxRunner.
+    """
+    if image_map is not None and not isinstance(image_map, dict):
+        raise TypeError(f"Expected image_map dictionary, got {type(image_map)}")
+    match runner_exec := runner.lower():
+        case "local":
+            use_local(**kwargs)
+        case "docker":
+            use_docker(
+                docker_executable=runner_exec,
+                image_overrides=image_map,
+                **kwargs,
+            )
+        case "singularity":
+            if not image_map:
+                raise ValueError("No container mappings provided.")
+            use_singularity(
+                singularity_executable=runner_exec, images=image_map, **kwargs
+            )
+        case _:
+            raise NotImplementedError(f"'{runner_exec}' runner not implemented.")
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,104 @@
+"""Tests for utility functions in neuromaps-prime."""
+
+from typing import Any
+
+import pytest
+from niwrap import DockerRunner, LocalRunner, SingularityRunner, get_global_runner
+
+from neuromaps_prime import utils
+
+
+class TestSetRunner:
+    """Testing of set runner utility method."""
+
+    @pytest.mark.parametrize(
+        "runner, expected_runner",
+        [
+            ("local", LocalRunner),
+            ("docker", DockerRunner),
+            ("singularity", SingularityRunner),
+        ],
+    )
+    def test_set_runner(
+        self,
+        runner: str,
+        expected_runner: LocalRunner | DockerRunner | SingularityRunner,
+    ):
+        """Test setting runner."""
+        images = {"container1": "/path/to/container"} if runner == "singularity" else {}
+        utils.set_runner(runner=runner, image_map=images)
+        test_runner = get_global_runner()
+        assert isinstance(test_runner, expected_runner)
+
+    @pytest.mark.parametrize(
+        "runner, expected_runner",
+        [
+            ("LOCAL", LocalRunner),
+            ("LoCaL", LocalRunner),
+            ("DOCKER", DockerRunner),
+            ("doCKeR", DockerRunner),
+            ("SINGULARITY", SingularityRunner),
+            ("siNgUlaRiTY", SingularityRunner),
+        ],
+    )
+    def test_runner_case_insensitive(
+        self,
+        runner: str,
+        expected_runner: LocalRunner | DockerRunner | SingularityRunner,
+    ):
+        """Test case-insensitivty of setting runner."""
+        images = (
+            {"container1": "/path/to/container"}
+            if runner.lower() == "singularity"
+            else {}
+        )
+        utils.set_runner(runner=runner, image_map=images)
+        test_runner = get_global_runner()
+        assert isinstance(test_runner, expected_runner)
+
+    @pytest.mark.parametrize(
+        "runner, expected_runner",
+        [
+            ("local", LocalRunner),
+            ("docker", DockerRunner),
+            ("singularity", SingularityRunner),
+        ],
+    )
+    def test_runner_valid_kwargs(
+        self,
+        runner: str,
+        expected_runner: LocalRunner | DockerRunner | SingularityRunner,
+    ):
+        """Test passing valid kwargs to set local runner."""
+        data_dir = "/path/to/data/dir"
+        images = {"container1": "/path/to/container"} if runner == "singularity" else {}
+        utils.set_runner(runner=runner, image_map=images, data_dir=data_dir)
+        test_runner = get_global_runner()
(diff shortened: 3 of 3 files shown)


src/neuromaps_nhp/datasets/data/surface_graph.yaml
kaitj: This looks great. 1 general comment overall and a note.

1. For at least sphere, pial, white, and midthickness, lets make sure those are consistent for keys across the nodes. For any of those that don't have a matching suffix, lets just double check to make sure those are correct. In general, we want these keys to consistent (unless it doesn't matter when you create the object).
2. This is the note: but to update the paths later on to the URL

surface_graph.png
kaitj: Looks great. I think this might be duplicated from the graph in `examples/`?

examples/test.py
kaitj: This seems to be an example script so I won't comment too much on the contents of it. That said, I would rename the file to reflect what it is doing or exclude it from the repository altogether (this kind of stuff will be good for documentation later on though, so maybe we want to keep it 😅)

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
...
+    file_path: Path
+
+    @abstractmethod
+    def fetch(self) -> Path:
+        """Fetch the resource."""
+        pass
+
+
+@dataclass
+class Surface_Atlas(Resource):
```
kaitj:
```suggestion
class SurfaceAtlas(Resource):
```

Pascal case for classes 😊

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
...
+        """Fetch the transform resource."""
+        return self.file_path
+
+
+class SurfaceGraph:
+    """Multi-directed graph of brain template spaces and their transformations."""
+
+    def __init__(self) -> None:
+        """Initialize an empty SurfaceGraph."""
+        self._spaces: Dict[str, List[Surface_Atlas]] = {}
```
kaitj: Thinking out loud here, does it make sense if the value is a `Set[SurfaceAtlas]`? Is there ever a situation where an atlas can show up multiple times within a key?

.pre-commit-config.yaml
nx10: Changes like bumping development dependency versions should typically be done in a separate PR unless they are related to the contents.

pyproject.toml
```diff
@@ -69,7 +76,7 @@ convention = "google"
 [tool.ruff.lint.per-file-ignores]
 "tests/**/*.py" = []
 
-[too.ruff.format]
+[tool.ruff.format]
```
nx10: Funny, although would put this in the same separate PR updating dev tooling as mentioned above

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
...
+    hemisphere: str
+    resource_type: str
+
+    def fetch(self) -> Path:
+        """Fetch the surface resource."""
+        return self.file_path
+
+
+@dataclass
+class Surface_Transform(Resource):
```
nx10: Python convention `SurfaceTransform`

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
+"""Module for managing brain template surface spaces and their transformations."""
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, Dict, List, Optional
```
nx10: Instead of `Dict`, `List`, `Option` use modern syntax (`list[str]`, `dict[str, Any]`, `str | None`)

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
...
+
+    source_space: str
+    target_space: str
+    density: str
+    hemisphere: str
+    resource_type: str
+
+    def fetch(self) -> Path:
+        """Fetch the transform resource."""
+        return self.file_path
```
nx10: Useless as long as file_path is a publicly accessible attribute: Either make it do actual I/O or remove abstraction

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
...
+        """Initialize an empty SurfaceGraph."""
+        self._spaces: Dict[str, List[Surface_Atlas]] = {}
+        self.graph = nx.MultiDiGraph()  # MultiDiGraph for multiple edges between nodes
+
+    def add_space(self, space_obj: Surface_Atlas) -> None:
+        """Add a space node to the graph."""
+        space_name = space_obj.space
+        if space_name not in self._spaces:
+            self._spaces[space_name] = []
+        self._spaces[space_name].append(space_obj)
```
nx10: Here surfaces are stored in `self._spaces` below they are duplicated in attributes. Pick one

src/neuromaps_nhp/graph.py
nx10: Consider adding `__repr__` methods to classes for easier debugging.

tests/test_graph.py
```diff
@@ -0,0 +1,64 @@
...
+        "neuromaps_nhp/datasets/data/surface_graph.yaml"
+    )
+    graph = graph_yaml(yaml_path)
+
+    def test_graph_yaml(self) -> None:
+        """Test loading the graph from YAML."""
+        assert self.graph is not None, "Graph loading failed"
+        assert self.graph.get_graph_info()["num_spaces"] == 6, "Expected 6 spaces"
+        assert self.graph.get_graph_info()["num_transforms"] == 48, (
+            "Expected 24 transforms (L and R hemispheres)"
```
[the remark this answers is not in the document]
nx10: Also the comment contradicts what is actually checked here ("== 48" vs "Expected 24")

examples/test.py
```diff
@@ -0,0 +1,66 @@
...
+target = "D99"
+if graph.has_path(source, target):
+    print(f"There is a path between {source} and {target}")
+else:
+    print(f"No path exists between {source} and {target}")
+print(f"shortest path = {graph.find_path(source, target)}")
+
+# Get information about Yerkes19 surfaces and transforms
+yerkes_surfaces = graph.get_spaces_by_name("Yerkes19")
+print(f"Yerkes19 has {len(yerkes_surfaces)} surface files")
```
nx10: I think duplicate lines to above

nx10 (commented):
Hey! Nice work getting this graph implementation together. The core architecture looks solid and you've built out a lot of useful functionality.

I've left inline comments on the specific issues, but wanted to summarize the main things to address:

- Class names like `Surface_Atlas` should be `SurfaceAtlas` (PEP 8 convention)
- Tests are sharing state with the class-level `graph` variable - better to use pytest fixtures so each test is isolated
- The `fetch()` method just returns the file path, so it's not really doing anything useful right now
- Test coverage is light - you're only testing about a third of the public methods
- Missing validation on things like hemisphere values and file existence
- As mentioned in the inline comments, the visualization logic could be cleaner in a separate module - it would make the core graph class easier to maintain and test
- The hardcoded `/home/bshrestha/...` paths in tests and examples will break for everyone else (although as I understand it Jason requested this for now)

Note that all of these are just suggestions and you can chat with Jason about what needs to be addressed before merging, in a follow-up PR, or not at all. Really nice job on the graph structure and the NetworkX integration - that's the hard part and you nailed it!

Also, heads up - my inline comments might seem terse, but that's just to keep things scannable. None of it is meant as criticism, just things to consider as you refine the code. Code review is always a collaborative process!

Let me know if you have questions about any of the feedback

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
...
+        """List all space names."""
+        return list(self.graph.nodes())
+
+    def get_graph_info(self) -> Dict[str, Any]:
+        """Get graph statistics."""
+        return {
+            "num_spaces": self.graph.number_of_nodes(),
+            "num_transforms": self.graph.number_of_edges(),
+            "spaces": list(self.graph.nodes()),
+            "is_connected": nx.is_weakly_connected(self.graph),
```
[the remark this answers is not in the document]
kaitj: It's not in our current mock-up, but I think could be useful to add as more nodes are added in the future as more get added on (either by us or on the fly by users).

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
...
+        for target in self.graph.successors(space_name):
+            edge_data = self.graph.get_edge_data(space_name, target)
+            for key, data in edge_data.items():
+                transforms["outgoing"].append(data["transform"])
+
+        # Incoming transforms
+        for source in self.graph.predecessors(space_name):
+            edge_data = self.graph.get_edge_data(source, space_name)
+            for key, data in edge_data.items():
+                transforms["incoming"].append(data["transform"])
```
kaitj: I don't believe we are planning on having the incoming transforms, just the outgoing? Will this be used anywhere / will things break if removed?

src/neuromaps_nhp/graph.py
```diff
@@ -0,0 +1,394 @@
...
+        except nx.NetworkXNoPath:
+            return []
+
+    def has_path(self, source: str, target: str) -> bool:
+        """Check if path exists between two spaces."""
+        result = nx.has_path(self.graph, source, target)
+        return bool(result)
+
+
+def graph_yaml(yaml_path: Path) -> SurfaceGraph:
```
kaitj: Just a note here (cc: @tamsinrogers) - will want to include methods for adding the volumetric stuff in here too. Once we have the `volume_graph.yaml` also included, I can combine them and either hard code that somewhere as a variable or just have a yaml file to load.

src/neuromaps_prime/plotting.py
kaitj: There are some optimizations that can be made within the functions of this module (e.g. single vs multiple implementation of code across conditions). Some of these will help performance and potentially make this easier to follow as well!

src/neuromaps_prime/graph.py
```diff
@@ -0,0 +1,603 @@
...
+    resolution: str
+    resource_type: str
+
+    def fetch(self) -> Path:
+        """Fetch the transform resource."""
+        return self.file_path
+
+    def __repr__(self) -> str:
+        """Custom string representation for debugging."""
+        return self.name
```
kaitj: Would move some of these methods into the pydantic models and remove these dataclasses - having both introduces redundancy.

src/neuromaps_prime/graph.py
```diff
@@ -0,0 +1,603 @@
...
+                f"Available spaces: {list(self.nodes)}"
+            )
+
+        node: Node = self.nodes[space]["data"]
+
+        for surface in node.surfaces:
+            if (
+                surface.density == validated_params.density
+                and surface.hemisphere.lower().startswith(
+                    validated_params.hemisphere.lower()[0]
```
kaitj: is the goal with this to match hemispheres by just matching the first letter?

kaitj (approved):
I think this is largely there - we can do the optimization bits later for both `graph.py` and `plotting.py` to push this through before release (just need to make an issue for it so we don't forget). Also need to consider what other validations we might need for the `models.py`.

PR #17, opened 2025-10-27
Add yaml schemas + examples

kaitj:
These were already living on a separate branch, this is just to bring it into `main`. The examples will eventually be replaced with the actual yaml files


birajstha (approved):
Just a small suggestion, feel free to ignore if not needed.
Otherwise LGTM :)

PR #19, opened 2025-10-28
Add niwrap runner utility method

kaitj:
This is just a simplified utility method I ported over and added tests from one of my other pipelines to more easily set the `StyxRunner` used. Put it under `neuromaps_prime.utils` to follow the `neuromaps` organization. Take a look and feel free to make any modifications as necessary. This is something that we will need to integrate down the line anyawys and make use of it is more robust.

### DIFF
--- a/src/neuromaps_prime/utils.py
+++ b/src/neuromaps_prime/utils.py
@@ -5,21 +5,21 @@ from niwrap import use_docker, use_local, use_singularity
 
 def set_runner(
     runner: str,
-    image_map: dict[str, str] = {},
+    image_map: dict[str, str] | None = None,
     **kwargs,
 ) -> None:
     """Set StyxRunner to use for NiWrap.
 
     Args:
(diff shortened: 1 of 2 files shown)


src/neuromaps_prime/utils.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    if not isinstance(image_map, dict):
+        raise TypeError(f"Expected image_map dictionary, got {type(image_map)}")
+    match runner_exec := runner.lower():
+        case "local":
+            use_local(**kwargs)
+        case "docker":
+            use_docker(
+                docker_executable=runner_exec,
+                image_overrides=image_map,
```
birajstha: should we pass an empty dict here? [Looks like it could be None](https://github.com/styx-api/styxdocker/blob/1862625afdf41da957a449f76a66c42a18180af4/src/styxdocker/__init__.py#L207C9-L207C24)

PR #25, opened 2025-10-30
Add pytest marker, runner selection capabilities

kaitj:
This PR primarily addresses the ability to select the Styx runner to be used for the testing suite, which can be useful when testing locally. As a benefit of this, certain tests can now be skipped if the required tool isn't installed / available (e.g. local runner on system without ANTs will skip all of the those tests that require ANTs). To support this two `pytest` specific arguments were added:

1. `runner`, which allows for the selection of the runner to be used: one of [`local`, `docker`, `singularity`]
2. `runner-images`, which is a mapping of images to overwrite docker (optional). or singularity (required) with

To use this, you can call pytest from the command line as such: 

```sh
# Use local runner
pytest --runner=local tests
# Use docker runner
pytest --runner=docker tests
# Use singularity runner
pytest --runner=singularity --runner-images={'antsx/ants:v2.5.3': 'docker://antsx/ants:v2.5.3'}
```

`transforms.py` and its tests were used to test this feature and as part of this I also added a missed test case for unsupported interpolators and moved some things around (i.e. if they functions were only used in the test, I moved it to the testing suite). As part of this, I also added a marker to check if ANTs was available for a specific testing suite. A `pytest.mark.usefixture('require_ants')` is used to decorate the suite (though this may be too broad of a coverage for this). A similar implementation will need to be added for workbench and any other tool we may have.

It would be helpful (until I can get the data_dir passing added) if either of you would be able to run this and see if it works with the current tests as is.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 """Global pytest fixtures, arguments, and options."""
 
+from pathlib import Path
 from typing import Generator
 
 import pytest
@@ -22,6 +23,12 @@ def pytest_addoption(parser: pytest.Parser):
         default=None,
         help="Optional JSON/dict string of Image overrides for Styx runner.",
     )
+    parser.addoption(
+        "--data-dir",
+        action="store",
+        default=None,
+        help="Directory where test data is located.",
+    )
 
 
 @pytest.fixture(scope="session", autouse=True)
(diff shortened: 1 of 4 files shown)


tests/transforms/test_volume.py
```diff
@@ -18,61 +18,82 @@
...
-    # Target images
-    target_same = Path(
-        "/Users/tamsin.rogers/Desktop/github/neuromaps/"
-        "share_with_T1w/Inputs/NMT2Sym/src-NMT2Sym_res-0p25mm_T1w.nii"
-    )
-    target_diff = Path(
-        "/Users/tamsin.rogers/Desktop/github/neuromaps/"
-        "share_with_T1w/Inputs/MEBRAINS/src-MEBRAINS_res-0p40mm_T1w.nii"
-    )
+    def _extract_res(self, nii_file: Path) -> tuple[float, float, float]:
```
birajstha: Does it make sense to move this in the transforms/utils?
Not sure if it will be used else where.

kaitj: Yeah, so this was in transforms/utils, but I pulled it into here because it wasn't being used anywhere but the test. I'm not sure if @tamsinrogers had plans to use it there? If so, then I agree it would make sense to move it back.

tests/conftest.py
```diff
@@ -0,0 +1,63 @@
+"""Global pytest fixtures, arguments, and options."""
```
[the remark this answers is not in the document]
kaitj: Yeah, this is a neat feature with `pytest` - would definitely recommend looking into this when building out a test suite!

birajstha (approved):

PR #26, opened 2025-10-30
Add argument to pass data directory to pytest CLI

kaitj:
This PR adds an argument `--data-dir` to the `pytest` CLI to allow for a slightly more dynamic file path to be passed to testing for now. This will allow one to use a local clone of the [`neuromaps-nhp-prep` directory ](https://github.com/tfunck/neuromaps-nhp
). To use, call `pytest --data-dir=<neuromaps-nhp-prep_dir> tests/`

Will need to make sure, at least for now, that the data directory file names are up to date and match across systems. Long term, we should just make use of the graph and its attributes once the files are added to the graph and available online.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 """Global pytest fixtures, arguments, and options."""
 
+from pathlib import Path
 from typing import Generator
 
 import pytest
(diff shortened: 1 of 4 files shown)


birajstha (approved):
This is awesome ! Thanks

tamsinrogers (approved):

PR #27, opened 2025-10-30
Surface to surface transformer

birajstha:
[Related to this issue .](https://github.com/tfunck/neuromaps-nhp-prep/issues/14)

This PR implements 
- [x] workbench.metric_resample
- [x] workbench.label_resample
- [x] surface to surface metric and label transformations 

- Currently the density to use in the intermediate space is chosen maximum available, for eg. Yerkes19 space has 32k and 10k surfaces so default 32k will be chosen. This can later me modified.
- Also, if no target_density is provided, the result metric/label will be resampled to the highest available density in the target space. 
- By default the new transforms computed will be added to the graph so, next time you want to do the transformations, this new transform will be used.


src/neuromaps_prime/transforms/utils.py
```diff
@@ -38,3 +40,62 @@ def get_vertex_count(surface_file: Path) -> int:
...
+    source: str,
+    target: str,
+) -> None:
+    """Validate that source and target spaces exist in the graph."""
+    if source not in graph.nodes(data=False) or target not in graph.nodes(data=False):
+        raise ValueError(
+            f"Either source space '{source}' or "
+            f"target space '{target}' does not exist in the graph."
+            f" Available spaces: {list(graph.nodes(data=False))}"
+        )
```
kaitj: I like the idea here - I think it is better to split this up to source AND target rather than so that we can alert the user to exactly what is missing.

src/neuromaps_prime/transforms/surface.py
```diff
@@ -53,3 +62,438 @@ def surface_sphere_project_unproject(
...
+    ----------
+    metric_in : Path
+        Path to input metric file.
+    current_sphere : Path
+        Path to current spherical surface.
+    new_sphere : Path
+        Path to new spherical surface.
+    metric_out : str
+        Path to output metric file.
+    method : str, optional
```
kaitj: Nice, for the final version we should enforce sampling methods. Also double check, but with niwrap, these resampling methods might be a workbench.metric_resample_param.

src/neuromaps_prime/transforms/surface.py
```diff
@@ -53,3 +62,438 @@ def surface_sphere_project_unproject(
...
+    """Generate hop output file path based on parameters."""
+    p = Path(output_file_path)
+    parent = p.parent
+    # Remove all suffixes for the base name
+    base = p.stem  # removes only the last suffix
+    # If the file has double suffix (e.g., .func.gii), remove both
+    if p.suffix == ".gii" and p.with_suffix("").suffix:
+        base = p.with_suffix("").stem
+    # Always use .surf.gii for intermediate sphere files
+    suffix = ".surf.gii"
```
kaitj: Should this not be `.sphere.gii` if it is a sphere - even if intermediate?

src/neuromaps_prime/graph.py
```diff
@@ -489,3 +488,88 @@ def get_node_data(self, node_name: str) -> Node:
...
+            surface_transforms=[surface_transform] if surface_transform else [],
+            volume_transforms=[volume_transform] if volume_transform else [],
+        )
+        self.add_edge(source_space, target_space, key=key, data=edge)
+
+    def search_surface_atlases(
+        self,
+        space: str,
+        density: str | None = None,
+        hemisphere: str | None = None,
```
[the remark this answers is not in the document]
birajstha: Its optional, in a sense - if None, will return all density and hemisphere data available.

src/neuromaps_prime/transforms/surface.py
```diff
@@ -53,3 +62,438 @@ def surface_sphere_project_unproject(
...
+        area_surfs=area_surfs,
+        method=method,
+    )
+    if not result.label_out.exists():
+        raise FileNotFoundError(f"Label out not found: {result.label_out}")
+
+    return result
+
+
+def _get_hop_output_file(
```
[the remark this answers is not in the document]
birajstha: its what came to my mind, hop [in a sense of network hopping nodes.](https://en.wikipedia.org/wiki/Hop_(networking))

src/neuromaps_prime/graph.py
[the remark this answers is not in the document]
birajstha: yes exactly

src/neuromaps_prime/transforms/surface.py
```diff
@@ -53,3 +62,438 @@ def surface_sphere_project_unproject(
...
+    p = Path(output_file_path)
+    parent = p.parent
+    # Remove all suffixes for the base name
+    base = p.stem  # removes only the last suffix
+    # If the file has double suffix (e.g., .func.gii), remove both
+    if p.suffix == ".gii" and p.with_suffix("").suffix:
+        base = p.with_suffix("").stem
+    # Always use .surf.gii for intermediate sphere files
+    suffix = ".surf.gii"
+    return str(
```
kaitj: Just trying to follow this pattern for the file name - could you provide a little more context for the naming here? If this is effectively naming a file for something like `1` -> `2` -> `3`, really just `1` -> `3`, then I think we should still follow the pattern's that are used in the prep repo. I am just way of having a long file name that may be difficult to interpret.

src/neuromaps_prime/transforms/utils.py
```diff
@@ -38,3 +40,62 @@ def get_vertex_count(surface_file: Path) -> int:
...
+
+
+def sort_density_key(d: str) -> int:
+    """Key function to sort density strings like '32k' numerically."""
+    return int(d.rstrip("kK")) if d.lower().endswith("k") else int(d)
+
+
+def find_common_density(
```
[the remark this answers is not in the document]
kaitj: Right, and if I remember right, with the downstream resampling, any interpolation that is necessary for mismatched densities will be performed automatically once it is in the right space.

src/neuromaps_prime/transforms/surface.py
```diff
@@ -53,3 +62,438 @@ def surface_sphere_project_unproject(
...
+
+    new_area_atlas = graph.fetch_surface_atlas(
+        space=target_space,
+        hemisphere=hemisphere,
+        density=target_density,
+        resource_type="midthickness",
+    )
+    if new_area_atlas is None:
+        raise ValueError(f"No midthickness surface found for {target_space}")
+    new_area: Path = new_area_atlas.fetch()
```
kaitj: The resource types used here should be a little more flexible - while we generally want to use the midthickness, other surfaces could be used in theory

src/neuromaps_prime/datasets/data/neuromaps_graph.yaml
```diff
@@ -445,13 +446,13 @@ edges:
...
         midthickness:
-          left: share/Outputs/Yerkes19-MEBRAINS/src-MEBRAINS_to-Yerkes19_den-101k_hemi-L_midthickness.surf.gii
-          right: share/Outputs/Yerkes19-MEBRAINS/src-MEBRAINS_to-Yerkes19_den-101k_hemi-R_midthickness.surf.gii
+          left: /home/bshrestha/projects/Tfunck/neuromaps-nhp-prep/share/Outputs/MEBRAINS-Yerkes19/src-yerkes19_to-MEBRAINS_den-32k_hemi-L_midthickness.surf.gii
+          right: /home/bshrestha/projects/Tfunck/neuromaps-nhp-prep/share/Outputs/MEBRAINS-Yerkes19/src-yerkes19_to-MEBRAINS_den-32k_hemi-R_midthickness.surf.gii
         sphere:
-          left: share/Outputs/Yerkes19-MEBRAINS/src-MEBRAINS_to-Yerkes19_den-101k_hemi-L_sphere.surf.gii
-          right: share/Outputs/Yerkes19-MEBRAINS/src-MEBRAINS_to-Yerkes19_den-101k_hemi-R_sphere.surf.gii
+          left: /home/bshrestha/projects/Tfunck/neuromaps-nhp-prep/share/Outputs/MEBRAINS-Yerkes19/src-yerkes19_to-MEBRAINS_den-32k_hemi-L_sphere.surf.gii
+          right: /home/bshrestha/projects/Tfunck/neuromaps-nhp-prep/share/Outputs/MEBRAINS-Yerkes19/src-yerkes19_to-MEBRAINS_den-32k_hemi-R_sphere.surf.gii
```
kaitj: Same for these paths (relative)

src/neuromaps_prime/graph.py
```diff
@@ -502,3 +761,262 @@ def get_node_data(self, node_name: str) -> Node:
...
+            raise ValueError(f"No atlases found for space '{space}'.")
+
+        highest_density = max(densities, key=_get_density_key)
+        print(f"Highest density for space '{space}': {highest_density}")
+        return highest_density
+
+    def surface_to_surface_transformer(
+        self,
+        transformer_type: str,
+        input_file: Path,
```
kaitj: Hah, this one threw me off. I thought it was asking them to provide sphere for some reason, but this makes sense.

kaitj (changes requested):
Largely looks good, thanks for moving methods to the `Graph` class. Just a minor thing with the paths in the yaml and the test paths.

src/neuromaps_prime/transforms/surface.py
```diff
@@ -53,3 +53,117 @@ def surface_sphere_project_unproject(
...
+        Path to new spherical surface.
+    output_file_path : str
+        Path to output label file.
+    **kwargs
+        Additional keyword arguments passed to workbench.label_resample.
+        Common options include:
+        - method : str
+            Resampling method. Default is 'ADAP_BARY_AREA'.
+        - area_surfs : tuple[Path, Path] | None
+            Tuple of area surfaces for adaptive barycentric resampling.
```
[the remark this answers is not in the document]
birajstha: @kaitj , Thanks for pointing this out. It does call the additional methods in the [actual code.](https://github.com/childmindresearch/neuromaps-prime/blob/2d0c2db9f6ed9f9bf9c4e927f1e31ac52887f438/src/neuromaps_prime/graph.py#L1001C9-L1003C10) Looks like these doc-strings are not correctly updated. 

Do we want to include other PARAMS too ? They are not used right now

src/neuromaps_prime/transforms/surface.py
```diff
@@ -53,3 +53,151 @@ def surface_sphere_project_unproject(
...
+
+    Raises:
+    ------
+    FileNotFoundError
+        If any input file does not exist.
+    """
+    current_sphere = kwargs.get("current_sphere")
+    new_sphere = kwargs.get("new_sphere")
+    method = kwargs.get("method")
+    area_surfs = kwargs.get("area_surfs")
```
kaitj: Similar comment as above

src/neuromaps_prime/datasets/data/neuromaps_graph.yaml
kaitj: Just wondering why `share` was removed from the paths?

kaitj (changes requested):
Almost there, a few minor things and then this one is good to merge!

src/neuromaps_prime/datasets/data/neuromaps_graph.yaml
kaitj: Ahh, i no i meant to move it to the parametrized arguments so we don't have to pass it within the test. Also maybe the yaml was updated since, I can't recall.

src/neuromaps_prime/graph.py
```diff
@@ -502,3 +761,265 @@ def get_node_data(self, node_name: str) -> Node:
...
+            resource_type=area_resource,
+        )
+        if new_area_atlas is None:
+            raise ValueError(f"No {area_resource} surface found for {target_space}")
+        new_area: Path = new_area_atlas.fetch()
+
+        kwargs = {
+            "current_sphere": transform.fetch(),
+            "new_sphere": new_sphere,
+            "method": "ADAP_BARY_AREA",
```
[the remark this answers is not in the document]
kaitj: Is this dict still being passed as `**kwargs`? If so then can add it as `"method": method`, otherwise no but maybe I am misunderstanding?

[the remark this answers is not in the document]
birajstha: @kaitj  Hmm, Since we are making these 3 inputs compulsory and skipping check inside the NiWrap wrapper function in surface  module, I think Its better to get rid of kwargs for now?

src/neuromaps_prime/datasets/data/neuromaps_graph.yaml
kaitj: I think some of these paths, or at least the way they are structured might be wrong?

For example, this looks right to me in terms of the directory structure and the filename.

```yaml
- from: MEBRAINS
    to: Yerkes19
    surfaces:
      101k:
        midthickness:
          left: share/Outputs/Yerkes19-MEBRAINS/src-MEBRAINS_to-Yerkes19_den-101k_hemi-L_midthickness.surf.gii
          right: share/Outputs/Yerkes19-MEBRAINS/src-MEBRAINS_to-Yerkes19_den-101k_hemi-R_midthickness.surf.gii
```

but this looks wrong to me:

```yaml
 - from: NMT2Sym
    to: Yerkes19
    surfaces:
      41k:
        midthickness:
          left: share/Outputs/NMT2Sym-Yerkes19/src-NMT2Sym_to-Yerkes19_den-41k_hemi-L_midthickness.surf.gii
          right: share/Outputs/NMT2Sym-Yerkes19/src-NMT2Sym_to-Yerkes19_den-41k_hemi-R_midthickness.surf.gii
        sphere:
          left: share/Outputs/NMT2Sym-Yerkes19/src-NMT2Sym_to-Yerkes19_den-41k_hemi-L_sphere.surf.gii
          right: share/Outputs/NMT2Sym-Yerkes19/src-NMT2Sym_to-Yerkes19_den-41k_hemi-R_sphere.surf.gii
```

Where the `source` and `target` are flipped in various places (key, directory, filename, etc.)

I think we probably need to sit down and go through this more thoroughly.

src/neuromaps_prime/transforms/surface.py
```diff
@@ -53,3 +53,132 @@ def surface_sphere_project_unproject(
...
 
     return result
+
+
+def metric_resample(
+    input_file_path: Path,
+    current_sphere: Path,
+    new_sphere: Path,
+    method: str,
+    area_surfs: workbench.metric_resample_area_surfs_params,
```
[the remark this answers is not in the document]
kaitj: Looks like it should be something like `workbench.MetricResampleAreaSurfsParameters`

Can take a look through this repo at this [specific commit](https://github.com/styx-api/niwrap-python/blob/85c631ff166df520954d595eb575d355c87a17bb/niwrap_workbench/src/niwrap_workbench/workbench/metric_resample.py)

src/neuromaps_prime/datasets/data/neuromaps_graph.yaml
```diff
@@ -151,46 +156,16 @@ nodes:
         flat:
           left: share/Inputs/S1200/src-S1200_den-32k_hemi-L_flat.surf.gii
           right: share/Inputs/S1200/src-S1200_den-32k_hemi-R_flat.surf.gii
+
 - CIVET:
```
kaitj: Let's just remove these altogether if we don't actually have data for them.

PR #31, opened 2025-11-04
Add `data_dir` argument to graph

kaitj:
This PR adds an additional argument to **optionally** allow a data directory to be passed when initializing the `Graph` object via the `data_dir` argument. For now, this is just used to point to the directory of where the data is stored. If provided, this serves as the parent-directory to the file paths provided in the yaml / dictionary object. The data directory can also be passed via the environment variable `NEUROMAPS_DATA`, which would be automatically pulled (though passing the argument will overwrite this).

The added test is incomplete, but ensures that the functionality is usable. Further, this doesn't break existing functionality. Future use case can make use of this variable to set the download location.

### DIFF
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -1,8 +1,24 @@
 """Test suite for graph functionalities in neuromaps_nhp."""
 
+from pathlib import Path
+
 import pytest
 
 from neuromaps_prime.graph import NeuromapsGraph
+
+
+@pytest.mark.usefixtures("require_data")
+def test_graph_initialization_with_data_dir(data_dir: Path, tmp_path: Path) -> None:
+    """Test initializing graph with data directory."""
+    graph = NeuromapsGraph(data_dir=data_dir)
+    assert graph is not None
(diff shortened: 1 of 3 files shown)


tests/test_graph.py
```diff
@@ -1,10 +1,25 @@
 """Test suite for graph functionalities in neuromaps_nhp."""
 
+from pathlib import Path
+
 import pytest
 
 from neuromaps_prime.graph import NeuromapsGraph
 
 
+def test_graph_initialization_with_data_dir(data_dir: Path, tmp_path: Path) -> None:
```
birajstha: do we need @pytest.fixture here?

tests/test_graph.py
```diff
@@ -1,10 +1,25 @@
 """Test suite for graph functionalities in neuromaps_nhp."""
 
+from pathlib import Path
+
 import pytest
 
 from neuromaps_prime.graph import NeuromapsGraph
```
birajstha:
```suggestion
@pytest.mark.usefixtures("require_data")
```

PR #32, opened 2025-11-05
Fix conditional doc build

kaitj:
Previous doc build runs were being skipped because either (a) tests were failing or (b) the condition for building the docs never triggered. This PR updates the condition to also trigger on a `workflow_dispatch` event.


birajstha (approved):

birajstha:
Yea, looks like the pages were also not enabled in the repo. I did enable them but feel free to change the config there.

PR #38, opened 2025-11-11
Reset runner after testing setting

### LINKED ISSUE #36: Bug: Pytest fails when running the whole tests dir
### Description

Pytest fails when running on the whole `tests` dir however, it runs fine when we give `tests/transforms` dir or individual files e.g. `test_graph.py` etc.

It gives out StyxSingularity related image not found error which hints its trying to use the Singularity Runner instead of LocalRunner specified.

In slack discussion, we have narrowed down to this file here which may have been leaking the runner state out and causing this failure.

### Steps to Reproduce

run tests on the higher level tests dir for e.g.
```BASH
uv run pytest tests --runner=local --data-dir /home/bshrestha/projects/Tfunck/neuromaps-nhp-prep
```

### Expected Results

Pytest passes when run on the whole tests dir

### Actual Results

Pytest is failing when running as a whole right now.

### Additional Notes

_No response_

kaitj:
Closes #36. This PR grabs the original runner before running the tests for setting different runner types and resets it after running the tests. There might be a more elegant solution to this, but this should get the failing tests working.

@birajstha, do you mind giving this a test?


birajstha (approved):
Yes, Tested on Local runner and it fixes the issue.

tamsinrogers (approved):

PR #57, opened 2025-12-02
Update models

kaitj:
- Moved `__repr__` from subclasses to the parent `Resource` class
- Switched `Node` and `Edge` classes to `pydantic` from `dataclass` (and moved this under models)
  - We initially planned on keeping these as `dataclass`, but `pydantic` would provide us the validation without too much additional work
- Set default weight of edges (transforms) to be 1.0
- Added unit testing for module.

### DIFF
--- a/src/neuromaps_prime/models.py
+++ b/src/neuromaps_prime/models.py
@@ -4,10 +4,10 @@ from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Literal
 
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, Field, field_validator
 
 
-class Resource(BaseModel, ABC):
+class Resource(BaseModel, ABC):  # pragma: no cover
     """Base model for resources in the neuromaps_prime graph."""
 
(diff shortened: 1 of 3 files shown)


src/neuromaps_prime/models.py
```diff
@@ -87,12 +99,48 @@ class VolumeTransform(Resource):
...
+            f"\tdescription={self.description}\n"
+            f"\tsurfaces=[{surface_str}]\n"
+            f"\tvolumes=[{volume_str}]"
+        )
+
+
+class Edge(BaseModel):
+    """Edge representation in transformation graph."""
+
+    model_config = ConfigDict()
```
birajstha: just out of curiosity, what does this do?

kaitj: Ahh, this is for `pydantic` if we want to introduce non default configuration for models (frozen attributes, arbitrary types, etc). For these models we actually don't need them since we aren't changing anything. 

It's probably better to remove this until we actually need it - thanks for catching!

Issue #29, opened 2025-11-03
Task: Add annotations to neuromaps_graph.yaml


tamsinrogers:
### Description

Add annotations to be made available as part of default graph.

### Tasks
- [x] Add object representation into code base. [d33d628](https://github.com/childmindresearch/neuromaps-prime/pull/96/commits/d33d62867494cd2c168ae29ca407c773a2c5a9ca) by @kaitj 
- [x] Add tests [d33d628](https://github.com/childmindresearch/neuromaps-prime/pull/96/commits/d33d62867494cd2c168ae29ca407c773a2c5a9ca) by @kaitj 
- [ ] Assess current annotations and define naming standards. Start with https://github.com/tfunck/neuromaps-nhp-prep/tree/main/resources/Yerkes19/annotations/receptor_maps.  Document [here](https://docs.google.com/document/d/1ADIR7dWmKrqQSlEEaPnb_QEMMSM_psbzwYJ5HhF3mjs/edit?tab=t.7d4mmsp1eyyx)
- [ ] Add annotation file paths to `neuromaps_graph.yaml`
- build off of the `refactor/graph` branch
- annotations should come one level after white/midthickness/etc, like this:
(https://github.com/childmindresearch/neuromaps-prime/blob/enh/graph-data/src/neuromaps_prime/resources/neuromaps_graph.yaml)
```
nodes:
- CIVETNMT:
    species: macaque
    description: CIVET Non-human Primate Template
    surfaces:
      41k:
        sphere:
          left: share/Inputs/CIVETNMT/src-CIVETNMT_den-41k_hemi-L_sphere.surf.gii
          right: share/Inputs/CIVETNMT/src-CIVETNMT_den-41k_hemi-R_sphere.surf.gii
        midthickness:
          left: share/Inputs/CIVETNMT/src-CIVETNMT_den-41k_hemi-L_midthickness.surf.gii
          right: share/Inputs/CIVETNMT/src-CIVETNMT_den-41k_hemi-R_midthickness.surf.gii
        white:
          left: share/Inputs/CIVETNMT/src-CIVETNMT_den-41k_hemi-L_white.surf.gii
          right: share/Inputs/CIVETNMT/src-CIVETNMT_den-41k_hemi-R_white.surf.gii
        annotation:
          receptormap: resources/CIVETNMT/receptor_maps
          anotherannotation: resources/CIVETNMT/anotherannotation
```

refer to
```
 "annotation": {
                                                "type": "object",
[...]

PR #96, opened 2026-03-10
Add keys for framework schema / model

### LINKED ISSUE #72: Task: Include "references" in schema and models
### Description

As we move towards adding annotations in (and even without), we should also include a `literature` or `citation` attribute both as part of the YAML schema and the model to allow for the inclusion or any relevant manuscripts and / or datasets. This should probably be a list to enable the inclusion of multiple items.

### Tasks

- [x] Update yaml schema
- [x] Update model

### Freeform Notes

_No response_

### LINKED ISSUE #82: Task: Add "Provider" Key
### Description

We want to be able to provide method for users to select the source of the transform (e.g. `CIVET` VS `Rhemap-Surf`)

### Tasks

- [x] Add key to YAML
- [x] Add support for selecting source in code

### Freeform Notes

Keys for references:
  - RheMap
  - CIVET
  - human_macaque

### LINKED ISSUE #95: Task: Add "notes" field
### Description

Stemming from recent discussions, we probably want to have something to add in warnings for any potential caveats specific to certain resources.

### Tasks

- [x] Look into how to best add this to schema
- [x] Implement in models

### Freeform Notes

This will likely look something similar to the literature, but also included for node resources (in addition to any annotations).

kaitj:
Closes #82 
- Updates the schema and configuration to support a `provider` key, allowing for selection of transform source
  - Current keys are placeholders using best guess and will need to be updated
  - Current behaviour is to use to the provided "key" as the preferred, but if not possible or not provided, it will use the first provider found as a fallback. At the moment, this is largely a thing for the multi-hop surfaces for cross-species.

Partially addresses #29 
- Updates schema to support `annotation` key, nested under volume and surface nodes
- Adds an `Annotation` model and methods into the codebase
- Updates unit tests for testing annotation 

Closes #72
- Updates schemas to support `references` key, which is a sequence of citations provided as a freeform string or object (in the YAML file)
- Add `references` attribute
- Update unit tests 

Closes #95
- Update schemas to support "notes", which is a sequence of freeform strings to provide additional information (e.g. caveats, limitations, etc.)
- Add `notes` attribute to model
  - _NOTE: This currently is supported only for annotations_
- Update unit tests  

Decided against adding fetchers for `notes` and `references`, as the annotations, transforms and surfaces can already be grabbed and the annotations can subsequently be grabbed from those attributes, rather than overengineering and creating unnecessary methods.

### DIFF
--- a/src/neuromaps_prime/graph/core.py
+++ b/src/neuromaps_prime/graph/core.py
@@ -362,6 +362,54 @@ class NeuromapsGraph(nx.MultiDiGraph):
             resource_type=resource_type,
         )
 
+    def search_volume_atlases(
+        self,
+        space: str,
+        resolution: str | None = None,
+        resource_type: str | None = None,
+    ) -> list[VolumeAtlas]:
+        """Search volume atlases for a space with optional filters.
+
+        Args:
+            space: Brain template space name.
+            resolution: Optional resolution filter.
+            hemisphere: Optional hemisphere filter.
+            resource_type: Optional resource type filter.
+
+        Returns:
+            List of matching :class:`~neuromaps_prime.graph.models.VolumeAtlas`
+            entries.
+        """
+        return self._cache.get_volume_atlases(
+            space=space, resolution=resolution, resource_type=resource_type
+        )
+
+    def search_volume_transforms(
+        self,
+        source_space: str,
+        target_space: str,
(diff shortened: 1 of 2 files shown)


kaitj:
Superceded by #123

PR #121, opened 2026-04-03
Split up testing

### LINKED ISSUE #73: Task: Add surface-to-volume transformer
### Description

Want to be able to transform from a surface annotation to volume.

### Tasks

- [x] Implement `surface_to_volume_transformer`

### Freeform Notes

Likely want to use `surface-to-surface` transforms as much as possible before switching to `volume-to-volume` transforms if necessary to complete transformation.

### LINKED ISSUE #97: Task: Split up testing
### Description

The current tests just put everything into a single test folder with appropriate `pytest` markers. While this works, I think we can better organize this. 

### Tasks

- [x] Split tests into `unit` and `integration`

### Freeform Notes

- `unit` tests should run without needing a container
- `integration` tests will call the container and also require real data

Possibly considering set up a self-hosted runner to run the integration tests to avoid having to pull containers using Github CI which has limited resources.

kaitj:
This splits up testing into 3 sensible modules - `unit`, `integration`, and `regression`. Some notable changes include:
  - adaptation of the `niwrap` module from [`rbc-mirror`](https://github.com/childmindresearch/rbc-mirror) for setting up the runners (future PR will need to deprecate this in the `utils` module and update the code base accordingly). This also closes out #108 
  - CI is marked to only run the `unit` tests - `integration` and `regression` should still be run prior to merging, which for now will have to be done locally as this requires both tools (e.g. workbench) being available and data

This PR also starts the introduction of stricter type-checking and formatting (at least via the `pre-commit` for now, and later on the entire codebase (via CI), with the deprecation of the `utils` module). `mypy` and `ruff` CIs may fail in the meantime until we get the codebase updated too (will be done in a future PR.


kaitj:
@tamsinrogers - I think this follows similar organization to how you started doing the surf cycle tests? If not, let me know!

tamsinrogers (dismissed):

kaitj:
Will rebase!

kaitj:
Merging in based on previous approval. Code base conflicted due to squash from another PR that this was based off, otherwise nothing has changed.

PR #134, opened 2026-04-30
Split YAML into multiple smaller yamls

### LINKED ISSUE #100: Task: Split YAML into multiple, smaller YAMLs
### Description

The current YAML is fairly large with multiple nodes and edges and will only grow bigger as more are added in along with different sources, making it different to edit. We should split these up into smaller YAMLs in a single resource directory that we can glob over and combine at initialization. Initial thought is:

```sh
resources
|- nodes
   |- macaque
      |- CIVETNMT
      |- ...
   |- human
      |- fsLR
      |- ...
|- edges
   |- macaque
   |- human
   |- xspecies
```

The directory name is mostly just to have some intuition of where things should go, but from a code perspective, globbing over these should be trivial.

### Tasks

- [ ] Split YAML
- [ ] Update code to combine on initliaization

### Freeform Notes

_No response_

kaitj:
This PR splits the one large (currently about 600 lines?) YAML file into multiple smaller YAML files that are more manageable, especially as more resources get added. 

- `resource` module is now organized by nodes and edges
  - nodes are further categorized by species and specific atlases within those species
  - edges are further categorized by type (surface vs volume) and species-to-species specific transformations
- schemas have been updated to handle these changes (along with `.vscode` settings to handle specific files)
  - CI updated to validate nodes and edges separately based on the updated schemas 
- codebase updated to handle the breakdown of multiple yaml files, merging them into a single dictionary before loading
- non-robust test for finding a common density updated to specific templates instead of just "first two in the graph"

Closes #100


kaitj:
@jpillai00 @tamsinrogers - will also need to move the annotations accordingly, but can do that in the necessary order depending on how PRs are merged

tamsinrogers (approved):

Issue #70, opened 2026-01-14
Task: Implement on-demand data download


tamsinrogers:
### Description

Ideally want to be able to fetch the static contents of `neuromaps-nhp-prep` and download on the fly as needed. A directory to download the data to will have to be specified (either user-provided or a chosen location).

### Tasks

- [x] Check API for OSF vs Zenodo (without login) for pulling data
- [ ] Move data to a host location (OSF, Zenodo, etc.) 
- [x] Add method for downloading data
- [x] Cache data to `DATA_DIR` so that it only needs to be downloaded if the file isn't found at the expected file path.

### Freeform Notes

- OSF
  - [API documentation](https://developer.osf.io/?ref=public_apis#tag/Versioning)
  - 5GB limit for private repositories / 50GB limit for public repositories.
- Zenodo
  - [API documentation](https://developers.zenodo.org/) 
  - 50GB limit per dataset

- Regardless of host location, potentially want multiple "projects" for the data that isn't already hosted elsewhere (maybe split by species?)

kaitj:
Completed via #142

PR #142, opened 2026-05-08
Enable remote data downloading

### LINKED ISSUE #149: Resource: Add neuromaps url
Add neuromaps nodes with osf urls for fetching

kaitj:
This PR adds the ability to download data from OSF remote storage and facilitates the ability to add on additional storage locations in the future. This also changes how the `data_dir` is used, from pointing to where the data is originally stored to the cache directory. This means that:

1. Any file paths provided in the YAML file are now assumed to be saved at the location provided. There is no longer an assumption that the file is saved in `self.data_dir / file_path`.  
2. Any data fetched from the remote servers are downloaded to this cache directory and neuromaps will handle this automatically

Data still needs to moved to OSF, would hold off on merging this until we have the data uploaded and / or the graphs updated to absolute paths / uris

_Partially addresses #70; still need to move the data_


tamsinrogers (approved):
looks great!

kaitj:
Merging this in. Only changes since previous approvals were merges from other PRs into this one.

PR #144, opened 2026-05-12
Surface Matrix Test

tamsinrogers:
**Surface Matrix Error Test** (tests/regression/test_surf_matrix.py)

_Compute a directed surface-to-surface transform error matrix by measuring median vertex-wise signed-distance error after resampling midthickness cortical surfaces between template spaces using sphere-based barycentric registration._

_The resulting matrix captures both directional (asymmetric) and symmetric geometric dissimilarity between brain surface atlases._
```
uv run pytest /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/test_surf_matrix.py -v -x -s --data-dir /Users/tamsin.rogers/Desktop/github/neuromaps-prime -o log_cli=true -o log_cli_level=DEBUG
```

```
=== TRANSFORM ERROR MATRIX ===
INFO     transforms.test_surf_matrix:test_surf_matrix.py:171 median surface-to-surface registration error (A → B)
INFO     transforms.test_surf_matrix:test_surf_matrix.py:172 
              fsLR   CIVETNMT        D99   MEBRAINS    NMT2Sym   Yerkes19
fsLR      0.000000  32.959244  32.494068  34.262169  32.517956  32.488098
CIVETNMT  5.823475   0.000000   0.358552   1.758073   5.664001   0.830897
D99       5.718718   0.409281   0.000000   1.774168   6.196968   0.880568
MEBRAINS  5.186776   1.348235   1.263126   0.000000   7.739522   1.287712
NMT2Sym   7.531346   5.008170   4.650307   7.188214   0.000000   4.494304
Yerkes19  4.869173   0.909239   0.854384   1.525678   6.231878   0.000000
```
Transform error matrix (directed): shows pairwise median surface registration error when mapping each source cortical space (rows) to each target space (columns) using sphere-based resampling, revealing overall compatibility structure across atlases.

```
=== ASYMMETRIC MATRIX ===
INFO     transforms.test_surf_matrix:test_surf_matrix.py:205 directionality bias in mapping; A → B vs B → A
INFO     transforms.test_surf_matrix:test_surf_matrix.py:206 
               fsLR   CIVETNMT        D99   MEBRAINS    NMT2Sym   Yerkes19
fsLR       0.000000  27.135768  26.775350  29.075393  24.986609  27.618926
CIVETNMT -27.135768   0.000000  -0.050729   0.409838   0.655831  -0.078342
D99      -26.775350   0.050729   0.000000   0.511043   1.546661   0.026183
MEBRAINS -29.075393  -0.409838  -0.511043   0.000000   0.551308  -0.237966
NMT2Sym  -24.986609  -0.655831  -1.546661  -0.551308   0.000000  -1.737575
Yerkes19 -27.618926   0.078342  -0.026183   0.237966   1.737575   0.000000
```
Asymmetry matrix: quantifies directional bias in registration by subtracting reverse mappings (A→B − B→A), highlighting strong non-reciprocity particularly for fsLR, where mapping into fsLR is substantially more distorted than mapping out.

```
=== SYMMETRIC MATRIX ===
INFO     transforms.test_surf_matrix:test_surf_matrix.py:210 average bidirectional transform error between spaces (A ↔ B)
INFO     transforms.test_surf_matrix:test_surf_matrix.py:211 
               fsLR   CIVETNMT        D99   MEBRAINS    NMT2Sym   Yerkes19
fsLR       0.000000  19.391360  19.106393  19.724473  20.024651  18.678635
CIVETNMT  19.391360   0.000000   0.383917   1.553154   5.336086   0.870068
D99       19.106393   0.383917   0.000000   1.518647   5.423637   0.867476
MEBRAINS  19.724473   1.553154   1.518647   0.000000   7.463868   1.406695
NMT2Sym   20.024651   5.336086   5.423637   7.463868   0.000000   5.363091
Yerkes19  18.678635   0.870068   0.867476   1.406695   5.363091   0.000000
```
Symmetric matrix: summarizes intrinsic geometric dissimilarity between atlas spaces by averaging bidirectional errors ((A→B + B→A)/2), revealing a coherent low-distortion cluster among CIVETNMT, D99, MEBRAINS, NMT2Sym, and Yerkes19, with fsLR as a clear outlier.

```
=== OFF-DIAGONAL TRANSFORM ERROR STATS ===
INFO     transforms.test_surf_matrix:test_surf_matrix.py:259 global median: 4.93867
INFO     transforms.test_surf_matrix:test_surf_matrix.py:261 global min: 0.35855 (CIVETNMT → D99)
INFO     transforms.test_surf_matrix:test_surf_matrix.py:262 global max: 34.26217 (fsLR → MEBRAINS)
INFO     transforms.test_surf_matrix:test_surf_matrix.py:264 NHP-only median: 1.64188
INFO     transforms.test_surf_matrix:test_surf_matrix.py:265 Human↔NHP median: 20.00972
```

```
Saved CSV → /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/outputs/surface_transform_matrix.csv
```

```
INFO     transforms.test_surf_matrix:test_surf_matrix.py:324 Saved full-scale heatmap → /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/outputs/surface_transform_matrix_full.png
INFO     transforms.test_surf_matrix:test_surf_matrix.py:354 Saved NHP-scaled heatmap → /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/outputs/surface_transform_matrix_nhp_scaled.png
```
image
image

The full-scale heatmap shows the complete range of transform errors across all atlas spaces, including fsLR, which allows for a  direct comparison of absolute magnitude differences but is visually dominated by high-error outliers.  The NHP-scaled heatmap uses a rescaled color limit based on the NPPatlas subset, suppressing fsLR  extremes to reveal relative differences within NHP.

```
INFO     transforms.test_surf_matrix:test_surf_matrix.py:369 Saved global histogram → /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/outputs/surface_transform_histogram.png
```
image
work in progress?


tests/integration/transforms/outputs/surface_transform_matrix_nhp_scaled.png
kaitj, line 1: For the heatmaps, it may also may sense to include the distances directly on each "cell".

tests/regression/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
+"""Compute direct surface-to-surface transform error matrix."""
+
+import logging
+from itertools import product
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import nibabel as nib
+import numpy as np
+import pandas as pd
```
kaitj, line 11: Make `pandas` an explicit dependency in the `pyproject.toml`. As it is used exclusively for the tests, I would add create a `test` dependency group (or `dev`).

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+    gii = nib.load(metric_file)
+    data = np.abs(gii.darrays[0].data)
+    return float(np.median(data))
+
+
+def fetch_surface(
+    graph: NeuromapsGraph,
+    space: str,
+    density: str,
+    hemi: str,
```
kaitj: Nitpicking, but the type hint for hemi should probably be something llike `Literal["left", "right"]`

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+def median_abs_signed_distance(metric_file: Path) -> float:
+    """Compute median absolute signed-distance error from func.gii."""
+    gii = nib.load(metric_file)
+    data = np.abs(gii.darrays[0].data)
+    return float(np.median(data))
+
+
+def fetch_surface(
```
kaitj: I think I understand the reasoning for creating a wrapper around our fetch method, but is it necessary?

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+            )
+
+            if sphere is not None and midthickness is not None:
+                valid.append(node)
+
+        except Exception as e:
+            logger.debug("Skipping node %s due to error: %s", node, e)
+            continue
+
+    return sorted(valid)
```
kaitj: How are the nodes sorted - is it by name?

tests/regression/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+
+    Each entry measures:
+        A_midthickness → B_midthickness
+    using sphere-defined barycentric mapping (triangle coordiantes).
+    """
+    logging.basicConfig(level=logging.INFO)
+
+    graph = NeuromapsGraph(
+        runner="local",
+        data_dir=Path("/Users/tamsin.rogers/Desktop/github/neuromaps-prime"),
```
kaitj: These should be left as the default for merging so that others can also run the test without necessarily having to change the `runner` or `data_dir` explicitly. PR #142 will help with dealing with the data location once everything is finalized there.

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+
+    hemisphere = "right"
+    spaces = get_valid_spaces(graph, hemisphere)
+
+    results = {}
+
+    logger.info("=== BUILDING SURFACE TRANSFORM MATRIX ===")
+
+    for src, dst in product(spaces, spaces):
+        if src == dst:
```
kaitj: Does workbench error out in this case? If not, it would be good to just let it calculate the distance as a sanity check.

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+        # so we use it for resampling
+        src_sphere = fetch_surface(graph, src, src_density, hemisphere, "sphere")
+        dst_sphere = fetch_surface(graph, dst, dst_density, hemisphere, "sphere")
+
+        # apply transform
+        out_surface = tmp_path / f"{src}_to_{dst}.surf.gii"
+
+        area_surfs = {
+            "current-area": str(src_surface),
+            "new-area": str(dst_surface),
```
kaitj: Nitpicking, but is string casting necessary?

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        # resample because the surfaces are not in the same space
```
kaitj: I think this comment means to say not the same mesh (e.g. number of vertices)?

tests/regression/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+        # compute error between resampled surface and target surface
+        error_file = tmp_path / f"{src}_to_{dst}_error.func.gii"
+
+        # now compute vertex-wise signed distance from the
+        # resampled surface to the target surface
+        workbench.signed_distance_to_surface(
+            surface_comp=str(out_surface),
+            surface_ref=str(dst_surface),
+            metric=str(error_file),
+        )
```
kaitj, line 179: Similar comment about casting to string as above.

tests/regression/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+            "current-area": str(src_surface),
+            "new-area": str(dst_surface),
+        }
+
+        # resample because the surfaces are not in the same space
+        workbench.surface_resample(
+            surface_in=str(src_surface),
+            current_sphere=str(src_sphere),
+            new_sphere=str(dst_sphere),
+            method="ADAP_BARY_AREA",
```
kaitj, line 165: string cast nitpick

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+        workbench.signed_distance_to_surface(
+            surface_comp=str(out_surface),
+            surface_ref=str(dst_surface),
+            metric=str(error_file),
+        )
+
+        # the absolute signed distance gives us a measure of how far the resampled
+        # surface is from the target surface at each vertex
+        # the sign indicates the direction of error (inside vs outside)
+        error = median_abs_signed_distance(error_file)
```
kaitj: Would also be interesting to report a mean and standard deviation here I think (or maybe a histogram of the errors to gauge how much vertices have moved)

tests/regression/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+    - Best representation of intrinsic atlas similarity
+    """
+
+    # assess symmetry
+    asymmetry = matrix - matrix.T
+    logger.info("\n=== ASYMMETRIC MATRIX ===")
+    logger.info("directionality bias in mapping; A → B vs B → A")
+    logger.info("\n%s", asymmetry)
+
+    symmetric = (matrix + matrix.T) / 2.0
```
kaitj, line 254: I think its in the docstrings, but this would be more like an average transformation error between two nodes?

tests/regression/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+    3. SYMMETRIC MATRIX
+    S[A, B] = (M[A, B] + M[B, A]) / 2
+
+    Interpretation:
+    - Undirected “geometric distance” between atlas spaces
+    - Best representation of intrinsic atlas similarity
+    """
+
+    # assess symmetry
+    asymmetry = matrix - matrix.T
```
kaitj, line 249: And this is more like a difference in transformation error going from one node to the other vs flipped direction?

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+    symmetric = (matrix + matrix.T) / 2.0
+    logger.info("\n=== SYMMETRIC MATRIX ===")
+    logger.info("intrinsic geometric difference between spaces")
+    logger.info("\n%s", symmetric)
+
+    # assess global error
+    off_diag = matrix.to_numpy()[~np.eye(len(matrix), dtype=bool)]
+    median_error = np.nanmedian(off_diag)
+
+    logger.info("Global median off-diagonal error: %.5f", median_error)
```
kaitj: I'm not sure if this is important to note since we are comparing different transformations between different nodes. What might be interesting (excluding the diagonals of course), is reporting on the min and max (potentially also separating the human from the NHPs since that will always be highest?) and what nodes those are.

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,267 @@
...
+    # CSV EXPORT
+    csv_path = output_dir / "surface_transform_matrix.csv"
+    matrix.to_csv(csv_path)
+    logger.info("Saved CSV → %s", csv_path)
+
+    # HEATMAP EXPORT
+    mat = matrix.to_numpy()
+    n = len(spaces)
+
+    # FULL-SCALE HEATMAP
```
kaitj: Noted in one of the plots, but would be nice to annotate the plots (similar to [this](https://matplotlib.org/stable/gallery/images_contours_and_fields/image_annotated_heatmap.html))

kaitj (commented):
Going to split this into parts:

1. Technical
    - Are we considering these to be integration tests (that is the functions we are calling work with real data) or are we planning for these to be part of regression testing (to see if adding / changing features change performance)? If the former, I would leave out all of the outputs proposed to be merged in here and share those offline or in a different repository. If the latter, I would leave out the matrices, but retain the .csv file and perhaps give the output directory a more meaningful name. In either case, I would update the `.gitignore` file appropriately.
    - A few type casting nitpicks
    - Explicitly adding a dependency that is being called (`pandas`)
2. Scientific (we can also take this discussion offline if it is easier)    
    - At a glance, the distances seem greater than I would have expected, in particular between the different macaque templates. This may just be a result of the transforms that we are being used, but we should double check to make sure we aren't doing anything unexpected. In any case, we've at least provided a framework now that we can perform these tests and substitute in other values.
    - Would also be interesting to have histograms of the distance errors from workbench to gauge how much the vertices have moved

Overall, I think it looks good, just a few changes to be made.

tfunck (commented):
I think it might make more sense to only to the surface mesh comparison within a given species. I'll check with Ting, but I don't think her cross-species transformation is really meant to produce a human-like surface mesh from a macaque one. 

I think NMT2Sym should have better results, so that might indicate that there is a mistake along the way.

tests/regression/test_surf_matrix.py
tamsinrogers: the default graph is still including `S1200` when the local directory is not set, so I think this needs some work
``
FAILED transforms/test_surf_matrix.py::test_surface_transform_matrix - FileNotFoundError: File path does not exist: share/Inputs/S1200/src-S1200_den-10k_hemi-L_sphere.surf.gii
``

.gitignore
```diff
@@ -51,6 +51,9 @@ coverage.xml
 .pytest_cache/
 cover/
 
+# Integration test outputs
+/tests/integration/transforms/surface_matrix_outputs/
```
kaitj: Leave out the leading "/", but looks fine otherwise. This also ignores this specific path; if you want to ignore all `surface_matrix_outputs` directories irrespective of where it is located in the repo (in case output_dir changes), you can remove all the parent directories, leaving only `surface_matrix_outputs/`

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dev = [
   "types-pyyaml>=6.0.12.20250915"
 ]
 docs = ["pdoc>=15.0.4"]
+tests = ["pandas>=3.0.3"]
```
kaitj, line 40: 👍 since you added this into a separate test group, would also update the [CI](https://github.com/childmindresearch/neuromaps-prime/blob/main/.github/workflows/test.yaml) to install and use the dependencies in this group. 

We don't currently call any of the integration or regression tests, but we may in the near future once we have the remote downloading merged in / have a dedicated self-hosted runner.

tests/regression/test_surf_matrix.py
kaitj, line 1: Thinking about this file more, I think this is probably a better fit as a regression test. We can reuse this to quantify performance, and further, with integration testing, we would be testing the functionality of the functions directly.

tests/integration/transforms/test_surf_matrix.py
```diff
@@ -0,0 +1,385 @@
...
+        vmin=vmin,
+        vmax=vmax,
+    )
+    annotate_heatmap(ax2, mat)
+
+    ax2.set_xticks(range(n))
+    ax2.set_yticks(range(n))
+    ax2.set_xticklabels(spaces, rotation=45, ha="right")
+    ax2.set_yticklabels(spaces)
+    ax2.set_title("Surface Transform Error Matrix (NHP-Scaled View; fsLR excluded)")
```
kaitj: Looking at the plots, its not so much that fsLR is excluded, but fsLR is clipped

kaitj (changes requested):
Looking good (code-wise). A few minor things to change, and looking at this more, this is a better fit as a regression test

tests/regression/test_surf_matrix.py
kaitj: If you are running the test locally, then this is expected since you do need to pass the data directory at the moment. On the CI (and if others clone this repository), this path wouldn't make sense. The alternative to this (and I can't recall at the moment if this was already implemented or not), but to use a environment variable on the local machine to pass the data directory so that we won't need to include this in the code. 

All of that being said, this problem would also be fixed in #142. What I would recommend doing for now is removing this path in the file for this PR, but keep it in the file locally (without pushing it).

.github/workflows/test.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - uses: astral-sh/setup-uv@v8.1.0
+    - run: uv sync --group tests
```
kaitj: Nice, would either add `--frozen` here or set the environment variable `UV_FROZEN=1`, so to ensure the `uv.lock` doesn't get updated (versions used for testing are what we expect)

kaitj (commented):
Almost, there. There is also still the set path [here](https://github.com/childmindresearch/neuromaps-prime/pull/144/changes#diff-dec4dec135e8ccd7e2fa3bf7821bdfaa802b5051c6a178d31983565dace182fcR98) for the graph initialization.

Other than that, it might also make sense to wait and potentially incorporate any feedback / changes stemming from @tfunck's [previous comment](https://github.com/childmindresearch/neuromaps-prime/pull/144#pullrequestreview-4283252867).

tfunck:
Thinking about the results of the human to macaque transformation, I think there's a good chance there's a mistake in the human -> macaque transformation because the errors are so much higher than for macaque -> human. Granted the matrix may not be symmetric but I think the error level should be similar between two spaces. Is it possible the wrong transformation file was used for human -> macaque?

EDIT: Actually, the above reasoning might not be correct. Ting's xspecies alignment optimizes the functional similarity between the macaque and human brain based on fMRI and doesn't care at all about morphology. If for example a human brain has an extra sulcus that the macaque doesn't, the alignment doesn't try to flatten out the human sulcus when fitting to the macaque. The macaque brains, by contrast, are aligned based on cortical curvature and sulcal depth which is explitly morphological.  So I guess it could be possible that the transformation is much worse in one direction versus the other. However, in either case, it doesn't make sense to use the vertex distance (which is a measure of morphological alignment) on transformations that aren't meant to optimize this. So like I was saying before, _we should only apply this test within species where morphological transformations have been calculated_.

tamsinrogers:
@tfunck here is some log output related to your above comment-

**fsLR → Yerkes19**
```
src_surface:
/Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/fsLR/src-fsLR_den-32k_hemi-R_midthickness.surf.gii
dst_surface:
/Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_midthickness.surf.gii
src_sphere:
/Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/S1200/src-S1200_den-32k_hemi-R_sphere.surf.gii
dst_sphere:
/Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_sphere.surf.gii

median=32.48810 mean=29.88887 std=14.22251
```

**Yerkes19 → fsLR**
```
src_surface:
/Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_midthickness.surf.gii
dst_surface:
/Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/fsLR/src-fsLR_den-32k_hemi-R_midthickness.surf.gii
src_sphere:
/Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_sphere.surf.gii
dst_sphere:
/Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/S1200/src-S1200_den-32k_hemi-R_sphere.surf.gii
median=4.86917 mean=5.94540 std=4.60708
```

So the large error does seem to be symmetry-related.  

I inspected the relevant files here:
```
(.venv) (base) CMI-RSCH-MBP118:neuromaps-prime tamsin.rogers$ wb_command -file-information /Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_sphere.surf.gii
Name:                       /Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_sphere.surf.gii
Type:                       Surface
Structure:                  CortexRight 
Data Size:                  1.17 Megabytes
Maps to Surface:            true
Maps to Volume:             false
Maps with LabelTable:       false
Maps with Palette:          false
Number of Vertices:         32492
Number of Triangles:        64980
Normal Vectors Correct:     true
Surface Type (Primary):     Spherical
Surface Type (Secondary):   Invalid
X-minimum:                  -99.995
X-maximum:                  99.995
Y-minimum:                  -99.995
Y-maximum:                  99.995
Z-minimum:                  -99.995
Z-maximum:                  99.995
Spherical Radius:           99.995
Surface Area:               125657.219
Spacing Mean:               2.122
Spacing Std Dev:            0.111
Spacing Minimum:            1.907
Spacing Maximum:            2.365

(.venv) (base) CMI-RSCH-MBP118:neuromaps-prime tamsin.rogers$ wb_command -file-information /Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/S1200/src-S1200_den-32k_hemi-R_sphere.surf.gii
Name:                       /Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/S1200/src-S1200_den-32k_hemi-R_sphere.surf.gii
Type:                       Surface
Structure:                  CortexRight 
Data Size:                  1.17 Megabytes
Maps to Surface:            true
Maps to Volume:             false
Maps with LabelTable:       false
Maps with Palette:          false
Number of Vertices:         32492
Number of Triangles:        64980
Normal Vectors Correct:     true
Surface Type (Primary):     Spherical
Surface Type (Secondary):   Invalid
X-minimum:                  -99.995
X-maximum:                  99.995
Y-minimum:                  -99.995
Y-maximum:                  99.995
Z-minimum:                  -99.995
Z-maximum:                  99.995
Spherical Radius:           99.995
Surface Area:               125657.219
Spacing Mean:               2.122
Spacing Std Dev:            0.111
Spacing Minimum:            1.907
Spacing Maximum:            2.365

(.venv) (base) CMI-RSCH-MBP118:neuromaps-prime tamsin.rogers$ wb_command -file-information /Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_midthickness.surf.gii
Name:                       /Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_midthickness.surf.gii
Type:                       Surface
Structure:                  CortexRight 
Data Size:                  1.17 Megabytes
Maps to Surface:            true
Maps to Volume:             false
Maps with LabelTable:       false
Maps with Palette:          false
Number of Vertices:         32492
Number of Triangles:        64980
Normal Vectors Correct:     true
Surface Type (Primary):     Anatomical
Surface Type (Secondary):   Midthickness
X-minimum:                  -1.265
X-maximum:                  29.351
Y-minimum:                  -46.124
Y-maximum:                  27.280
Z-minimum:                  -17.396
Z-maximum:                  25.466
Surface Area:               11604.327
Spacing Mean:               0.681
Spacing Std Dev:            0.209
Spacing Minimum:            0.149
Spacing Maximum:            1.787

(.venv) (base) CMI-RSCH-MBP118:neuromaps-prime tamsin.rogers$ wb_command -file-information /Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_sphere.surf.gii
Name:                       /Users/tamsin.rogers/Desktop/github/neuromaps-prime/share/Inputs/Yerkes19/src-Yerkes19_den-32k_hemi-R_sphere.surf.gii
Type:                       Surface
Structure:                  CortexRight 
Data Size:                  1.17 Megabytes
Maps to Surface:            true
Maps to Volume:             false
Maps with LabelTable:       false
Maps with Palette:          false
Number of Vertices:         32492
Number of Triangles:        64980
Normal Vectors Correct:     true
Surface Type (Primary):     Spherical
Surface Type (Secondary):   Invalid
X-minimum:                  -99.995
X-maximum:                  99.995
Y-minimum:                  -99.995
Y-maximum:                  99.995
Z-minimum:                  -99.995
Z-maximum:                  99.995
Spherical Radius:           99.995
Surface Area:               125657.219
Spacing Mean:               2.122
Spacing Std Dev:            0.111
Spacing Minimum:            1.907
Spacing Maximum:            2.365
```

does anything stand out to you or @kaitj ?

kaitj:
Nothing at a glance, but if the alignment is based on function rather than morphology than I think it makes sense for the large errors. We should then make a note in the graph edge about this if this is indeed the case.

tfunck:
It could be useful to visualize the transformed surfaces versus their targets. It'll give us a sense of what's going on.

kaitj (commented):
From a code perspective, I think this is largely good to merge in since a lot of the regression testing will require a user to identify noted changes.

There are a few things from a CI perspective that might make sense to do (if this can ultimately be run on a self-hosted runner, the github runners may be too limited), is to:

1. Save the csv and images to the repository for tags, and 
2. Automatically run and compare results against the previous tag + upload the images for each PR (similar to what is done [here (see artifacts)](https://github.com/childmindresearch/rbc/actions/runs/26307474360)).

@tfunck anything from a testing perspective that you think we might be missing?

tfunck:
I don't think so on my end. I agree with your suggestions related to CI.

PR #146, opened 2026-05-21
add NMT2Sym, D99, and CIVETNMT annotation entries to YAML

jpillai00:
Updates the annotations for the NMT2Sym, D99, and CIVETNMT yaml files based on changes to groupings/names [here](https://docs.google.com/spreadsheets/d/1EDaRNpxcC8TRUmbzrc4X5GX9i0EjFqf5qn1oJ4A437c/edit?gid=1645453905#gid=1645453905).

To-Do:
- [x] update CIVETNMT CHARM files annots
- [x] remove AT as an annotation category & replace with PC for NMT2Sym, D99

*will address updating 5 more files in NMT2Sym in a different PR


src/neuromaps_prime/resources/nodes/macaque/D99.yaml
```diff
@@ -15,6 +15,27 @@ D99:
       white:
         left: https://files.osf.io/v1/resources/nmxfh/providers/osfstorage/69d52303cc9143f354902958?view_only=b214f40525fe44b5bea3103478d08d0d
         right: https://files.osf.io/v1/resources/nmxfh/providers/osfstorage/69d523022882bdb2bc3ebdee?view_only=b214f40525fe44b5bea3103478d08d0d
+
   volumes:
     250um:
       T1w: https://files.osf.io/v1/resources/nmxfh/providers/osfstorage/69d52302a52f9bb3897129cf?view_only=b214f40525fe44b5bea3103478d08d0d
+      annotation:
+        PC:
```
kaitj, line 23: What is different about this one versus the one below (PC_sym)?

src/neuromaps_prime/resources/nodes/macaque/CIVETNMT.yaml
```diff
@@ -12,3 +12,14 @@ CIVETNMT:
       white:
         left: https://files.osf.io/v1/resources/nmxfh/providers/osfstorage/69d522eda52f9bb3897129b8?view_only=b214f40525fe44b5bea3103478d08d0d
         right: https://files.osf.io/v1/resources/nmxfh/providers/osfstorage/69d522ec4f3931c54c90480e?view_only=b214f40525fe44b5bea3103478d08d0d
+      annotation:
```
kaitj: uri's for the file(s) themselves? i am assuming this is just a text file mapping label to structure.

src/neuromaps_prime/resources/nodes/macaque/D99.yaml
```diff
@@ -15,6 +15,27 @@ D99:
...
+            - 'lookup file: https://files.us.osf.io/v1/resources/nmxfh/providers/osfstorage/6a0f510b7da771710ca7aa44?view_only=b214f40525fe44b5bea3103478d08d0d'
+        PC_sym:
+          uri: https://files.us.osf.io/v1/resources/nmxfh/providers/osfstorage/6a0f510a270229e9b9679c4d?view_only=b214f40525fe44b5bea3103478d08d0d
+          notes:
+            - Symmetrized/mirrored variant of the D99 v2.0 atlas.
+        PC_hemi_R:
+          uri: https://files.us.osf.io/v1/resources/nmxfh/providers/osfstorage/6a0f510c45728813016799d7?view_only=b214f40525fe44b5bea3103478d08d0d
+          notes:
+            - Right-hemisphere-only subset of the D99 v2.0 atlas.
+        PC_seg_cortical:
```
kaitj: uri or does it use an existing uri (which we should include here as well anyways if we have a notes).

kaitj (commented):
Every annotation should probably be associated with a uri unless there is a good reason not to. Otherwise, see notes / questions below.

src/neuromaps_prime/resources/nodes/macaque/CIVETNMT.yaml
jpillai00: only these text files were dropped into that folder

src/neuromaps_prime/resources/nodes/macaque/D99.yaml
jpillai00: did not look into this but i'll open up the files and check!

src/neuromaps_prime/resources/nodes/macaque/D99.yaml
jpillai00: same as the CIVETNMT comment, but I believe only this text file was in the original D99 folder. i will double check

src/neuromaps_prime/resources/nodes/macaque/CIVETNMT.yaml
kaitj: If these text files aren't associated with any specific annotation or something specific to the space, would leave them out. It wouldn't really make sense to include them unless we / a user knows what to do with them. What do the contents of these files look like?

src/neuromaps_prime/resources/nodes/macaque/D99.yaml
kaitj: It's possible that this is using the D99 parcellation and instead is just a look up table, in which case i would include this in the notes there.

jpillai00: yes, I believe it is using the D99 parcellation. here is the file: [src-D99_seg-cortical_desc-PC_annot.txt](https://github.com/user-attachments/files/30092975/src-D99_seg-cortical_desc-PC_annot.txt)

I already have noted down that it is a lookup file in the notes:`cortical-only lookup file:` do I need to include anything more?

src/neuromaps_prime/resources/nodes/macaque/D99.yaml
jpillai00: they look exactly the same when I opened it up so I will remove the `PC_sym` version.

src/neuromaps_prime/resources/nodes/macaque/CIVETNMT.yaml
jpillai00: that makes sense, I can remove them.

here are those files if helpful:

[CHARM_atlas_rsl_sym_hires_level_3.txt](https://github.com/user-attachments/files/30093161/CHARM_atlas_rsl_sym_hires_level_3.txt)
[CHARM_atlas_rsl_sym_hires_level_4.txt](https://github.com/user-attachments/files/30093162/CHARM_atlas_rsl_sym_hires_level_4.txt)
[macaque_mid_mc_left.txt](https://github.com/user-attachments/files/30093160/macaque_mid_mc_left.txt)

kaitj: Ahh I see, they are associated with the CHARM atlas (which I think is also ~D99~ NMT2Sym?) and I think a mean curvature (if I am interpretting the mc correctly)? In this case, it probably makes sense to remove them.

src/neuromaps_prime/resources/nodes/macaque/D99.yaml
kaitj: This is the PC nifti above right? In that case, I would remove this individual key and move the note to the nifti above just so they are associated with the same key.

src/neuromaps_prime/resources/nodes/macaque/NMT2Sym.yaml
```diff
@@ -15,6 +15,53 @@ NMT2Sym:
       inflated:
         left: https://files.osf.io/v1/resources/nmxfh/providers/osfstorage/69d523dfab76e18c73904626?view_only=b214f40525fe44b5bea3103478d08d0d
         right: https://files.osf.io/v1/resources/nmxfh/providers/osfstorage/69d523df4d366de13d714041?view_only=b214f40525fe44b5bea3103478d08d0d
+      annotation:
+        SD:
+          right: https://files.us.osf.io/v1/resources/nmxfh/providers/osfstorage/6a0f5213cd87cca253679ac7?view_only=b214f40525fe44b5bea3103478d08d0d
+          notes: 
+            - Sulcal depth map (hemi-R)
```
kaitj: Could consider leaving this note out since the keys should already be indicative of what this represents. I think you already mentioned this, but we have a mapping of the acronym (e.g. SD) already.

kaitj (commented):
Looks fine to me now, just waiting for the YAML validation CI to pass before approving.

kaitj:
Hmm, a bunch of urls (unrelated to this PR) are failing due from lack of access to the files. @tamsinrogers, we should double check the project ids + view only keys on these. I'll also try to update the script so that we aren't getting rate limited.

@jpillai00, going to try rerunning the ci one more time in a few hours, but otherwise this looks good to go. Doesn't seem like the failing uris are related to the annotations added.

kaitj (approved):
Alright, all but 2 are passing now and for both of those, I tried to manually check the the uri and was having trouble loading osf. May be overloaded or rate limited.

PR #147, opened 2026-05-21
replace local paths with OSF

tamsinrogers:
@kaitj I'm not sure this is exactly correct, trying to follow along with #142. Let me know if the paths should be structured differently!


kaitj:
> @kaitj I'm not sure this is exactly correct, trying to follow along with #142. Let me know if the paths should be structured differently!

It's not quite right. You'll have to use the view only link (passing the token) and grab it via the osf api. I had an example of this somewhere at one point, I'll see if I can find it and send it to you!

kaitj:
Superceded by #154, closing.

PR #152, opened 2026-05-27
Docs: resource GH templates Fix #102

### LINKED ISSUE #102: Task: Add template(s) for adding resource
### Description

Tied closely with #101, we should also add in some pre-defined GitHub template for requesting or contributing a new resource.

### Tasks

- [ ] Add PR template
- [ ] Add Issue template

### Freeform Notes

_No response_

jpillai00:
- [X] Added resource request issue template at `.github/ISSUE_TEMPLATE/resource_request.yaml`
- [x] Added general PR template at `.github/pull_request_template.md`

Closes #102


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Resource Contribution (if applicable)
+**Resource Type:** (Annotation / Atlas / Transform / Other)
+
+**Resource URI:**
+
+**References:**
+
+**YAML Changes:**
+<!-- Which YAML file(s) were updated and under which resource category? -->
```
kaitj: I'm not sure we need these details to be included in the PR - these would show up in the PR diff anyways. This should definitely be checked by the review though.

.github/ISSUE_TEMPLATE/resource_request.yaml
kaitj, line 1: I think this may be a little too specific on the details. This to me seems like it would be good if we expect a single issuer per resource, but if someone wants to contribute an entire dataset that seems like it would be a lot of issues. I would probably generalize this more instead of singular dropdowns (maybe checkboxes). I guess I am primarily thinking of `resource-type`

kaitj (changes requested):
I think its a great start, really just a minor change I would make.

kaitj (approved):
lgtm 🚀. I'm basing this is off of the yaml completely, so if something is off once merged in we can make a patch fix PR

PR #162, opened 2026-06-03
fix plotting bug for surface and volume subgraphs

jpillai00:
## This PR:
Fixes a bug in `plotting.py` where `_plot_single_graph` was missing parameters passed from `plot_graph`, causing an error when plotting single surface or volume subgraphs.

```
TypeError: _plot_single_graph() got an unexpected keyword argument 'legend_rect'
```

## Type of Change
- [X] Bug fix
- [ ] Enhancement
- [ ] Resource contribution
- [ ] Documentation
- [ ] Other

### DIFF
--- a/src/neuromaps_prime/plotting.py
+++ b/src/neuromaps_prime/plotting.py
@@ -264,12 +264,17 @@ def _plot_single_graph(
     save_path: Path | None,
     layout: str,
     colormap: str,
+    legend_rect: tuple[float, float, float, float],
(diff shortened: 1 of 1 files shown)


kaitj (approved):
Not too particular about the arguments for the plotting here, but lgtm 🚀. 

I think it was left out of the `plot_single_graph` because it was never explicitly passed, but is part of the `**kwargs`

PR #163, opened 2026-06-04
Added graph example

gkiar:
## This PR:

Adds a plotting script that turns the graph into a nice html visual that we can use as the basis for presentations and the like. Note, this also includes the svg export patch that Flo vibe coded (adapted by my own vibe coding, ofc).

## Type of Change
- [ ] Bug fix
- [ ] Enhancement
- [ ] Resource contribution
- [x] Documentation
- [ ] Other

P.s., this is downstream of branch `enh/new-nodes` but I think that branch may no longer exist and thus please direct me as to where we should target the PR.


kaitj:
The branch has been merged into `main`, so a rebase from that branch should do the trick. It was also squash merged so I think only the last 2 commits here are needed.

gkiar:
Ah ok, since it was a squash, it may be cleaner for me to do a new branch and cherry-pick the last two commits?

kaitj:
Sure, that works!

kaitj:
I just went ahead and cleaned up the commit history + pointed it to merge into `main`. Otherwise, looks good to me!

PR #166, opened 2026-06-05
Fix volume transform atlas lookup & file extension bug

jpillai00:
## This PR:
Fixes for bugs when running volume transforms from running end-to-end. All were found while testing the example volume transform script with Yerkes19 -> MEBRAINS transforms.

 1. Volume atlas lookup fails when transform and atlas have different resource types.`volume_to_volume_transformer` used a single `resource_type` parameter for both the transform lookup and the reference atlas lookup. The MEBRAINS to Yerkes19  transform is registered as `composite` (a warp field) while the Yerkes19 reference  atlas is registered as `T1w`, so no single value worked for both lookups.
 - **Error:** `ValueError: No volume atlas found for 'Yerkes19' (resolution='500um', resource_type='composite')`
  - **Fix:** Added `atlas_resource_type` param that defaults to "T1w" so the transform lookup uses `resource_type`, the atlas lookup uses `atlas_resource_type`. 

2. Volume atlas and transform file paths missing `.nii `extension so ANTs failed when trying to use the ref atlas without extension
- **Error:** `file /styx_input/1/Yerkes19_500um_T1w is not recognized as a supported image format antsApplyTransforms: Segmentation fault`
-  **Fix:** Changed `file_path` construction to `data_dir / f"{name}.nii.gz"` for both volume resources and volume annotations. `name` field kept without extension because it is used to identify the file

## Type of Change
- [X] Bug fix
- [ ] Enhancement
- [ ] Resource contribution
- [ ] Documentation
- [ ] Other

## Related Issue(s)

### DIFF
--- a/src/neuromaps_prime/graph/core.py
+++ b/src/neuromaps_prime/graph/core.py
@@ -640,7 +640,8 @@ class NeuromapsGraph(nx.MultiDiGraph):
             output_file_path: Path for the warped output volume.
             interp: Interpolation method.
             interp_params: Optional interpolation parameters.
-            atlas_resource_type: Volume resource type for the reference atlas lookup (default ``'T1w'``).
+            atlas_resource_type: Volume resource type for the reference atlas lookup
+                (default ``'T1w'``).
             provider: Optional provider name. Falls back to the first
                 registered provider when ``None``.
 
--- a/src/neuromaps_prime/graph/core.py
+++ b/src/neuromaps_prime/graph/core.py
@@ -636,12 +636,12 @@ class NeuromapsGraph(nx.MultiDiGraph):
             source_space: Source brain template space.
             target_space: Target brain template space.
             resolution: Target volume resolution (e.g. ``'2mm'``).
-            resource_type: Volume resource type (e.g. ``'composite'``).
+            resource_type: Volume transform resource type (e.g. ``'composite'``).
             output_file_path: Path for the warped output volume.
             interp: Interpolation method.
             interp_params: Optional interpolation parameters.
-            atlas_resource_type: Volume resource type for the reference atlas lookup
-                (default ``'T1w'``).
+            atlas_resource_type: Volume resource type for the reference atlas 
+                lookup (default ``'T1w'``).
             provider: Optional provider name. Falls back to the first
                 registered provider when ``None``.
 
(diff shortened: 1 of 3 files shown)


src/neuromaps_prime/graph/transforms/volume.py
```diff
@@ -70,10 +71,11 @@ def transform_volume(
             source_space: Source brain template space name.
             target_space: Target brain template space name.
             resolution: Target volume resolution (e.g. ``'1mm'``, ``'500um'``).
-            resource_type: Volume resource type (e.g. ``'T1w'``, ``'composite'``).
+            resource_type: Volume resource type (e.g. ``'composite'``).
```
kaitj:
```suggestion
            resource_type: Volume transform resource type (e.g. ``'composite'``).
```

src/neuromaps_prime/graph/transforms/volume.py
```diff
@@ -70,10 +71,11 @@ def transform_volume(
...
             output_file_path: Path for the warped output volume.
             interp: Interpolation method passed to the warp tool.
             interp_params: Optional additional interpolation parameters.
+            atlas_resource_type: Volume resource type for the reference atlas lookup (default ``'T1w'``).
```
kaitj:
```suggestion
            atlas_resource_type: Volume atlas resource type for the reference atlas lookup (default ``'T1w'``).
```

src/neuromaps_prime/graph/core.py
```diff
@@ -635,10 +636,11 @@ def volume_to_volume_transformer(
             source_space: Source brain template space.
             target_space: Target brain template space.
             resolution: Target volume resolution (e.g. ``'2mm'``).
-            resource_type: Volume resource type (e.g. ``'T1w'``).
+            resource_type: Volume resource type (e.g. ``'composite'``).
```
kaitj:
```suggestion
            resource_type: Volume transform resource type (e.g. ``'composite'``).
```

src/neuromaps_prime/graph/core.py
```diff
@@ -635,10 +636,11 @@ def volume_to_volume_transformer(
...
             output_file_path: Path for the warped output volume.
             interp: Interpolation method.
             interp_params: Optional interpolation parameters.
+            atlas_resource_type: Volume resource type for the reference atlas lookup (default ``'T1w'``).
```
kaitj:
```suggestion
            atlas_resource_type: Volume atlas resource type for the reference atlas lookup (default ``'T1w'``).
```

kaitj (changes requested):
A couple of docstring changes, but looks good to me otherwise.

kaitj (approved):

PR #170, opened 2026-06-11
add current graph image

tamsinrogers:


kaitj (changes requested):
Looks good in the location, but the plot itself is very small. Do we have a vectorized version of this (or bigger)?

(requesting for a bigger version 😄)

kaitj (approved):

Issue #175, opened 2026-06-12
Task: update receptor map OSF pointers


tamsinrogers:
### Description

Our OSF project is now public. Unreleased data has been moved to Neuromaps-PRIME-private.

### Tasks

- [x] remove the receptor maps from `annotations` in OSF
- [x] make current project public
- [x] re-upload the receptor maps in a private project
- [ ] fix the receptor map OSF pointers
- [ ] optionally remove the view only key from all other OSF pointers

PR #176, opened 2026-06-12
update receptor map pointers

tamsinrogers:
## This PR:
Updates the OSF receptor map pointers now that they are stored in a different project

## Type of Change
- [x] Bug fix
- [ ] Enhancement
- [ ] Resource contribution
- [ ] Documentation
- [ ] Other

## Related Issue(s)
- #175 by @tamsinrogers


kaitj:
The URLs are not correct, which is why the CI for validating the graphs is failing. You'll have to update the project id, as well as the token associated with that project. The current project id and token is associated with the now public OSF repo so won't work with the new one.

kaitj (changes requested):
See previous comment about the url / project id.

tamsinrogers:
@kaitj oops! fixed them

kaitj (approved):
Thanks, lgtm 🚀The new pointers are working.

PR #183, opened 2026-06-24
Experimental transformations warning

kaitj:
## This PR:
Adds an experimental warning for certain transformations. Also includes a `provider` tag to filter for specific transformations to warn for. The assumption is that both a forward and reverse transform would be experimental, so only one needs to be provided. 

## Type of Change
- [ ] Bug fix
- [x] Enhancement
- [ ] Resource contribution
- [ ] Documentation
- [ ] Other

## Related Issue(s)
n/a

### DIFF
--- a/src/neuromaps_prime/graph/transforms/surface.py
+++ b/src/neuromaps_prime/graph/transforms/surface.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import logging
 from pathlib import Path
-from typing import Any, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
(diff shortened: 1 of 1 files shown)


kaitj:
I couldn't remember when building this, but I think it was the MBM-NMT2Sym transformation (marmoset-macaque) that was experimental? This can easily be updated.

_cc: @tfunck (if you'd be able to verify which one was experimental?)_

tfunck:
Yep that's the one. The MBM is the marmoset brain atlas space

tamsinrogers (approved):

PR #198, opened 2026-07-17
Tf cycle

tfunck:
# PR Overview
This branch adds end-to-end cycle regression testing on the real Neuromaps graph to validate transform roundtrip quality across multi-hop paths.

#What’s included
1. New real-graph cycle regression tests
* Enumerate return paths from origin spaces
* Run hop-by-hop surface roundtrips
* Score path quality (Pearson r, max absolute difference)
2 . Artifact generation for debugging
* Per-run output folders with random suffixes
* CSV/TXT summaries and per-path transform manifests
* Surface plots and optional wb_command logs
3 Robust execution across real data variability
* Runs both left and right hemispheres
* Falls back for area surfaces in order:
** midthickness
** pial
** white
* Skips unusable paths/resources with clear warnings instead of hard-failing whole runs
Container-safe output handling
4. Adds an additional dependency, matplotlib_surface_plotting, for visual QC artifacts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,7 +40,10 @@ dev = [
   "types-requests>=2.33.0.20260518"
 ]
 docs = ["pdoc>=15.0.4"]
-tests = ["pandas>=3.0.3"]
+tests = [
+  "pandas>=3.0.3",
+  "matplotlib-surface-plotting>=0.1.0"
+]
 validate = ["check-jsonschema>=0.37.1"]
 
 [build-system]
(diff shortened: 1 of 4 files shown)


kaitj (commented):
🚀 Looks like a great starting point for us to make the tests more robust. I think there are a lot of things we can move to be test fixtures that can be re-used across the different testing modes (`unit`, `regression`, etc.). We'll put together a plan to tackle this "testing" milestone! 

There are some ruff CI complaints, but not too worried about those as we can tackle those as part of making the tests more robust.

kaitj:
Closing, superceded by #220

Issue #204, opened 2026-07-21
Task: update edge provider keys


tamsinrogers:
### Description

use `provider` key to actually describe where the transform is coming from in edge yamls

### Tasks

- [ ] for example: replace `human_chimpanzee` with `CIVET`, replace`macaque_marmoset` with `experimental_NeuromapsPRIME`

tamsinrogers:
closed by #205

Issue #203, opened 2026-07-21
Task: edge references


tamsinrogers:
### Description

edge references

### Tasks

- [ ] add references (to papers, datasets) in edge yamls (as in node yamls)

tamsinrogers:
closed by #205 #207

Issue #202, opened 2026-07-21
Task: add RheMap-Surf edges


tamsinrogers:
### Description

use `provider` key to add github path

### Tasks

- [ ] should look like:

```
- from: CIVET
  to: fsLR
  surfaces:
    neuromaps:
      41k:
        sphere:
          left: 
    other_source:
      41k:
        sphere:
          left: 

```

start with https://github.com/HumanBrainED/RheMAP-Surf/tree/main/deformation_spheres/standard_alignment

can indicate `RheMAPSurf_standard` and then do `RheMAPSurf_approxvolume`

tamsinrogers:
closed by #205

PR #205, opened 2026-07-21
update edge providers

tamsinrogers:
## This PR:
- updates provider tags to be more descriptive
- adds RheMAP-Surf edge pointers
- adds references to edge yamls

## Type of Change
- [ ] Bug fix
- [x] Enhancement
- [ ] Resource contribution
- [ ] Documentation
- [ ] Other

## Related Issue(s)
- #204 by @tamsinrogers 
- #203 by @tamsinrogers 
- #202 by @tamsinrogers

### DIFF
--- a/src/neuromaps_prime/resources/edges/surface/macaque_marmoset.yaml
+++ b/src/neuromaps_prime/resources/edges/surface/macaque_marmoset.yaml
@@ -1,7 +1,7 @@
 - from: MBM
   to: NMT2Sym
   surfaces:
-    macaque_marmoset:
+    experimental_NeuromapsPRIME:
       38k:
         sphere:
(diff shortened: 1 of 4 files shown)


src/neuromaps_prime/resources/edges/surface/macaque_marmoset.yaml
```diff
@@ -1,7 +1,7 @@
 - from: MBM
   to: NMT2Sym
   surfaces:
-    macaque_marmoset:
+    experimental_NeuromapsPRIME:
```
kaitj: We could leave out `experimental` here since we do have a separate variable keep track of experimental transformations within the graph itself that will throw a warning if experimental edges between two nodes are encountered

kaitj (changes requested):
Largely looks good - for the RheMap-Surf transformations, preference would be to use the tagged versions URLs instead of `main` as the files themselves could change on a new PR / push

kaitj (approved):

Issue #209, opened 2026-07-23
Task: [TESTING: Unit]


tamsinrogers:
### Unit Tests
GOAL: develop small, fast, isolated tests focused on validating individual components and graph operations.
- Add minimal cyclic transformation tests using synthetic graphs (e.g., 3 small nodes).
  - Verify that closed transformation cycles preserve data.
  - Confirm expected behavior (e.g., perfect correlation after completing a cycle).
- Refactor existing test implementations to improve maintainability:
  - Use pytest fixtures for shared setup and reusable test components.
  - Standardize test structure and styling.
  - Improve readability and organization of existing test files.

### Tasks

- [ ] add and refactor unit_test_cycle.py. use pytest fixtures, styling

kaitj:
Closed via #218

PR #218, opened 2026-07-27
Unit Cycle Test (refactor)

tamsinrogers:
## This PR:
addresses https://github.com/childmindresearch/neuromaps-prime/milestone/5
goal: small, cyclic test  with 3 small dummy nodes - should return perfect correlation

Added isolated unit tests for cycle transformations using a synthetic three-node graph. Tests verify cycle discovery, expected graph behavior, and that closed transformation paths preserve input metrics with near-perfect correlation. The Workbench resampling step is mocked to keep tests fast and reproducible.

**Updated 08/04 with clearer methods, fixtures, docstrings, and structure**

## Type of Change
- [ ] Bug fix
- [x] Enhancement
- [ ] Resource contribution
- [ ] Documentation
- [ ] Other

## Related Issue(s)
- #209  by @tamsinrogers 

**run test:**
```
uv run pytest /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/unit/graph/test_cycle.py -v
```
**current output:**
```
======================================================= test session starts =======================================================
platform darwin -- Python 3.12.2, pytest-9.1.1, pluggy-1.6.0 -- /Users/tamsin.rogers/Desktop/github/neuromaps-prime/.venv/bin/python3
cachedir: .pytest_cache
rootdir: /Users/tamsin.rogers/Desktop/github/neuromaps-prime
configfile: pyproject.toml
plugins: cov-7.1.0
collected 5 items                                                                                                                 

unit/graph/test_cycle_unit.py::test_find_return_paths_returns_expected_cycles PASSED
unit/graph/test_cycle_unit.py::test_find_return_paths_returns_all_simple_cycles PASSED
unit/graph/test_cycle_unit.py::test_find_return_paths_rejects_unknown_node PASSED
unit/graph/test_cycle_unit.py::test_closed_cycles_preserve_metric 
---------------------------------------------------------- live log call ----------------------------------------------------------
INFO     tests.cycle:cycle.py:333 cycle A -> B -> A: r=0.999998 max|delta|=1.552e-02
INFO     tests.cycle:cycle.py:333 cycle A -> C -> A: r=0.999998 max|delta|=1.781e-02
INFO     tests.cycle:cycle.py:333 cycle A -> B -> C -> A: r=0.999998 max|delta|=1.963e-02
INFO     tests.cycle:cycle.py:333 cycle A -> C -> B -> A: r=0.999999 max|delta|=2.211e-02
INFO     tests.cycle:cycle.py:346 Cycle test results — origin: A, hemisphere: left
INFO     tests.cycle:cycle.py:346 ------------------------------------------------------------------------------
INFO     tests.cycle:cycle.py:346 Transformation path                                  Pearson r     Max |delta|
INFO     tests.cycle:cycle.py:346 ------------------------------------------------------------------------------
INFO     tests.cycle:cycle.py:346 A -> B -> A                                           0.999998       1.552e-02
INFO     tests.cycle:cycle.py:346 A -> C -> A                                           0.999998       1.781e-02
INFO     tests.cycle:cycle.py:346 A -> B -> C -> A                                      0.999998       1.963e-02
INFO     tests.cycle:cycle.py:346 A -> C -> B -> A                                      0.999999       2.211e-02
INFO     tests.cycle:cycle.py:346 ------------------------------------------------------------------------------
INFO     tests.cycle:cycle.py:346 Total cycles: 4
PASSED
unit/graph/test_cycle_unit.py::test_concatenated_transform_matches_direct_transform PASSED

======================================================== 5 passed in 6.70s ========================================================
(base) Mac:tests tamsin.rogers$ 
```


tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+    from neuromaps_prime.graph import NeuromapsGraph
+
+logger = logging.getLogger(__name__)
+
+SURFACE_EDGE = "surface_to_surface"
+
+
+def resolve_hop_transforms(
+    graph: NeuromapsGraph,
+    path: tuple[str, ...],
```
kaitj: nit, but would rename this from `path`; makes me think we are providing a file path of some sort instead of the cyclic path

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+import networkx as nx
+import nibabel as nib
+import numpy as np
+
+if TYPE_CHECKING:
+    from neuromaps_prime.graph import NeuromapsGraph
+
+logger = logging.getLogger(__name__)
+
+SURFACE_EDGE = "surface_to_surface"
```
kaitj: Does this need to be defined again or is this already defined in the package and can be imported to be used?

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+        path: Cycle path, e.g. ``('A', 'B', 'C', 'A')``.
+        hemisphere: ``'left'`` or ``'right'``.
+        density: Preferred surface mesh density for the lookup.
+        edge_type: Graph edge layer to traverse (default ``'surface_to_surface'``).
+
+    Returns:
+        List of ``(source_space, target_space, density_used, file_path)``
+        tuples, one per direct graph edge, in traversal order.
+    """
+    rows: list[tuple[str, str, str, Path]] = []
```
kaitj: I think this may be better suited to be a `list[NamedTuple]` so that we can directly reference the attribute rather than having to remember that the last item in the tuple or the first item is the source space for example.

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+                    )
+                else:
+                    logger.warning(
+                        "No sphere transform found for %s -> %s (%s) at any density.",
+                        hop_src,
+                        hop_dst,
+                        hemisphere,
+                    )
+            if transform is not None:
+                rows.append((hop_src, hop_dst, density_used, transform.file_path))
```
kaitj: Given this is a loop, I think I would just add a `continue` to the else statement above and skip this conditional, which assumes that the data is available to append to the row.

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+    return rows
+
+
+def _path_token(path: tuple[str, ...]) -> str:
```
kaitj, line 51: Low collision risk for the returned token here for repeated calls of the same path. Why only take the first 12 values of the hex in the function and also why return a partially human-readable token here vs just returning the hash directly?

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+            if origin not in cycle:
+                continue
+            # Rotate the directed cycle so it starts at origin, then close it.
+            start = cycle.index(origin)
+            rotated = cycle[start:] + cycle[:start] + [origin]
+            paths.append(tuple(rotated))
+        return sorted(paths, key=lambda p: (len(p), p))
+
+    if max_length is None:
+        raise ValueError("max_length is required when allow_revisits=True")
```
kaitj: A minor nit, but would move this to the top so it is caught early.

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+
+    # Enumerate bounded round-trips built from two simple directed legs.
+    found: set[tuple[str, ...]] = set()
+    # Turning node cannot be origin, otherwise round-trip is length 0.
+    turn_nodes = [n for n in subgraph.nodes if n != origin]
+    for turn in turn_nodes:
```
kaitj: So this runs if `allow_revisits=True` but is a nested for-loop. Minor for the unit test where we have a small graph, but could be problematic for denser graphs. The `max_paths` variable isn't applied until after these loops are completed.

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+            "cycle %s: r=%.6f max|delta|=%.3e",
+            " -> ".join(path),
+            pearson_r,
+            max_abs_diff,
+        )
+        results.append(
+            CycleResult(path=path, pearson_r=pearson_r, max_abs_diff=max_abs_diff)
+        )
+
+    # Print summary table to stdout.
```
kaitj: Minor, but could de-duplicate this summary so that the write and the print can re-use a single method.

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+                # Fall back to any registered density for this edge.
+                candidates = graph.surface_ops.cache.get_surface_transforms(
+                    source=hop_src,
+                    target=hop_dst,
+                    hemisphere=hemisphere,
+                    resource_type="sphere",
+                )
+                if candidates:
+                    transform = candidates[0]
+                    density_used = transform.density
```
kaitj: Double check that the cache is sorted / non-deterministic, otherwise this will change every time it is run.

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+        origin: Space the metric starts from and must return to.
+        edge_type: Edge key to traverse (``'surface_to_surface'`` or
+            ``'volume_to_volume'``).
+        max_length: Optional cap on the number of edges in a path. ``None``
+            enumerates all simple cycles, which grows combinatorially on dense
+            graphs; bound it on the real graph. Required when
+            ``allow_revisits=True``.
+        allow_revisits: If ``True``, enumerate two-leg round-trips (outbound +
+            return simple paths) up to ``max_length``.
+        max_paths: Optional cap on number of returned paths. Useful with
```
kaitj: Double check this - I think number of returned paths here refers to edges? There might be an off by 1 if networkx is counting nodes.

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+            "Round-tripped metric did not return to the origin mesh: "
+            f"{roundtrip.shape} vs {original.shape}."
+        )
+    finite_mask = np.isfinite(original) & np.isfinite(roundtrip)
+    if np.any(finite_mask):
+        max_abs_diff = float(
+            np.max(np.abs(original[finite_mask] - roundtrip[finite_mask]))
+        )
+    else:
+        max_abs_diff = float("nan")
```
kaitj: There should be a warning raised if NaN is returned

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+    original_finite = original[finite_mask]
+    roundtrip_finite = roundtrip[finite_mask]
+    original_std = float(np.std(original_finite))
+    roundtrip_std = float(np.std(roundtrip_finite))
+    if np.isclose(original_std, 0.0) or np.isclose(roundtrip_std, 0.0):
+        pearson_r = 1.0 if vectors_equal else 0.0
+    else:
+        pearson_r = float(np.corrcoef(original_finite, roundtrip_finite)[0, 1])
+        if np.isnan(pearson_r) and vectors_equal:
+            pearson_r = 1.0
```
kaitj, line 391: If both vectors are NaN (unlikely, but not impossible), there can be a misleading perfect correlation.

tests/cycle.py
```diff
@@ -0,0 +1,473 @@
...
+
+if TYPE_CHECKING:
+    from neuromaps_prime.graph import NeuromapsGraph
+
+logger = logging.getLogger(__name__)
+
+SURFACE_EDGE = "surface_to_surface"
+
+
+def resolve_hop_transforms(
```
kaitj: Is this used anywhere? I also don't think the returned density is ever used in the tuple.

tests/cycle.py
kaitj, line 1: This provides a lot of the utility functions to be used I think, are there any fixtures to be added? Could also consider renaming this to provide more context. Can brain storm this a little bit later on.

tests/unit/graph/unit_test_cycle.py
kaitj, line 1: Might need to rename this for pytest to automatically pick it up as part of the full testing suite.

pyproject.toml
```diff
@@ -40,7 +40,10 @@ dev = [
   "types-requests>=2.33.0.20260518"
 ]
 docs = ["pdoc>=15.0.4"]
-tests = ["pandas>=3.0.3"]
+tests = [
+  "pandas>=3.0.3",
+  "matplotlib-surface-plotting>=0.1.0"
```
kaitj: I don't think this surface plotting dependency is being used currently.

tests/unit/graph/unit_test_cycle.py
```diff
@@ -0,0 +1,420 @@
...
+    image = nib.load(str(path))
+
+    for array in image.darrays:
+        if array.data.ndim == 2 and array.data.shape[1] == 3:
+            return np.asarray(array.data, dtype=float)
+
+    raise ValueError(f"No vertices found in {path}")
+
+
+def _load_metric(path: Path | str) -> np.ndarray:
```
kaitj: This is also repeated in the cycle utility module. Would define in one place and just import it.

tests/unit/graph/unit_test_cycle.py
```diff
@@ -0,0 +1,420 @@
...
+    """Load a vertex-wise metric from a GIFTI functional file."""
+    return np.asarray(
+        nib.load(str(path)).darrays[0].data,
```
kaitj: I don't think the path needs to be cast as a `str`

tests/unit/graph/unit_test_cycle.py
```diff
@@ -0,0 +1,420 @@
...
+    d00 = np.einsum("ij,ij->i", v0, v0)
+    d01 = np.einsum("ij,ij->i", v0, v1)
+    d11 = np.einsum("ij,ij->i", v1, v1)
+    d20 = np.einsum("ij,ij->i", v2, v0)
+    d21 = np.einsum("ij,ij->i", v2, v1)
+
+    denominator = d00 * d11 - d01 * d01
+    denominator = np.where(
+        np.abs(denominator) < 1e-18,
+        1e-18,
```
kaitj: Is `1e-18` a static threshold? Maybe we can just make it a constant?

tests/unit/graph/unit_test_cycle.py
```diff
@@ -0,0 +1,420 @@
...
+        weights,
+        metric[indices],
+    )
+
+
+def _fake_metric_resample(
+    input_file_path: str | Path,
+    current_sphere: str | Path,
+    new_sphere: str | Path,
+    method: str,  # noqa: ARG001
```
kaitj: what is the ARG001 for? Maybe we can add a comment if these are necessary to be included for `ruff`

tests/unit/graph/unit_test_cycle.py
```diff
@@ -0,0 +1,420 @@
...
+    return SimpleNamespace(metric_out=Path(output_file_path))
+
+
+# -------------------------------------------------------------------------
+# Fixtures
+# -------------------------------------------------------------------------
+
+
+@pytest.fixture
+def output_dir(tmp_path: Path) -> Path:
```
kaitj: Is this fixture scoped per test or per module (can't recall what the default is at the moment) - I think it might be per test. Also is a fixed output directory necessary for the unit testing vs creating them per test as needed?

tests/unit/graph/unit_test_cycle.py
```diff
@@ -0,0 +1,420 @@
...
+
+def test_find_return_paths_returns_expected_cycles(
+    rotation_network: tuple[NeuromapsGraph, Path],
+) -> None:
+    """Return all simple closed paths from the synthetic graph."""
+    graph, _ = rotation_network
+
+    paths = find_return_paths(graph, "A")
+
+    assert set(paths) == EXPECTED_CYCLES
```
kaitj, line 411: Is this equality deterministic? I don't think `sets` are ordered (I might be wrong) and wondering if that is a factor in the assertion here.

(Also for some of the assertions in the below tests)

kaitj (changes requested):
Largely looks good. There are a few questions, nits, and bugs with silent failures / behavior. 

With fixtures, thinking broader (can any of them be used for regression testing for example).

tests/unit/graph/unit_test_cycle.py
tamsinrogers: Correct, sets aren't ordered so this is deterministic. So I think we just need to assert that the cycles exist in no particular order?

kaitj: Yes. Just double checked, didn't realize `EXPECTED_CYCLES` was also a set.

tests/unit/graph/test_cycle_unit.py
```diff
@@ -0,0 +1,528 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from types import SimpleNamespace
+
+import nibabel as nib
+import numpy as np
+import pytest
+from nibabel.gifti import GiftiDataArray, GiftiImage
+from scipy.spatial import ConvexHull, cKDTree
```
kaitj: I think `cKDTree` has started to throw a warning that it will be deprecated in an upcoming version and to just use `KDTree` - also see [note](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.cKDTree.html) in their docs.

tests/unit/graph/test_cycle_unit.py
```diff
@@ -0,0 +1,528 @@
...
+HEMISPHERE = "left"
+
+EXPECTED_CYCLES = {
+    ("A", "B", "A"),
+    ("A", "C", "A"),
+    ("A", "B", "C", "A"),
+    ("A", "C", "B", "A"),
+}
+
+DENOM = 0.00000000001
```
kaitj: Nit, but would make more pythonic: `1e-11` in this case

tests/unit/graph/test_cycle.py
```diff
@@ -0,0 +1,528 @@
...
+        [
+            [1, 0, 0],
+            [0, np.cos(theta), -np.sin(theta)],
+            [0, np.sin(theta), np.cos(theta)],
+        ],
+        dtype=np.float64,
+    )
+
+
+def _save_surface(
```
kaitj, line 119: Just noting that these save and load functions feel like they may be useful for other unit tests. Fine to leave here for now, but we may want to pull them out to their own test helper module in the future (or if they end up elsewhere in the package to just import them)

tests/unit/graph/test_cycle_unit.py
```diff
@@ -0,0 +1,528 @@
...
+# -------------------------------------------------------------------------
+# Tests
+# -------------------------------------------------------------------------
+
+
+def test_find_return_paths_returns_expected_cycles(
+    rotation_graph: tuple[NeuromapsGraph, Path],
+) -> None:
+    """Find all simple closed paths originating from the synthetic graph node A."""
+    graph = rotation_graph  # retrieve graph
```
kaitj: This isn't needed - the graph is already passed and `rotation_graph` can be called directly.

tests/unit/graph/test_cycle_unit.py
```diff
@@ -0,0 +1,528 @@
...
+) -> None:
+    """Verify that identity transformation cycles preserve a vertex-wise metric.
+
+    Each discovered cycle propagates the synthetic metric through the surface
+    transformations and back to the starting space. Since the synthetic
+    transformations compose to identity, the returned metric should closely
+    match the original metric.
+    """
+    graph = rotation_graph
+    metric = rotation_metric
```
kaitj: Already passed in. Generally unless there is some mutation being performed to the inputs, there is no need to assign new variables.

tests/unit/graph/test_cycle_unit.py
```diff
@@ -0,0 +1,528 @@
...
+        metric,
+        HEMISPHERE,
+        workdir,
+        density=DENSITY,
+    )
+
+    assert {result.path for result in results} == EXPECTED_CYCLES
+
+    for result in results:
+        assert result.pearson_r > 0.999
```
kaitj: Would make use `pytest.approx` to perform this comparison with an absolute difference tolerance. Something like this I think should work:

```suggest
    assert result.pearon_r == pytest.approx(1.0, abs=1e-3)
```

tests/unit/graph/test_cycle_unit.py
```diff
@@ -0,0 +1,528 @@
...
+        density=DENSITY,
+    )
+
+    pearson_r, max_abs_diff = score_roundtrip(
+        direct,
+        concatenated,
+    )
+
+    assert pearson_r > 0.999
+    assert max_abs_diff < 0.02
```
kaitj: Here to, I would look to use `pytest.approx`

tests/unit/graph/test_cycle.py
```diff
@@ -0,0 +1,528 @@
...
+    preserving the expected resampling function interface.
+    """
+    monkeypatch.setattr(
+        "neuromaps_prime.graph.transforms.surface.metric_resample",
+        _fake_metric_resample,
+    )
+
+
+# -------------------------------------------------------------------------
+# Tests
```
kaitj, line 401: A very minor nit and not necessary to change, but general preference towards using test classes to group together similar tests (e.g. return paths).

tests/cycle.py
kaitj, line 1: Just noting this as I also started to write some helpers on for the stats unit tests - we can make a decision later. I wonder if its helpful to organize it like `tests.helpers.<module>` (or maybe even `tests.unit.helpers` since they would mostly all be unit test associated)?

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+                if counts.get(origin, 0) != 2:
+                    continue
+
+                if any(node != origin and count > 2 for node, count in counts.items()):
+                    continue
+
+                yield roundtrip
+
+
+def find_return_paths(
```
kaitj, line 138: I would consider making everything after `graph` (or possibly `origin`) a keyword-argument (moving the `*`).

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+    graph: NeuromapsGraph,
+    origin: str,
+    edge_type: str = NeuromapsGraph.surface_to_surface_key,
+    *,
+    max_length: int | None = None,
```
kaitj, line 143: I think I am a little confused, would you be able to explain the difference between `max_length` and `max_path`? Is `max_length` the longest possible path and `max_path` the number of possible paths (e.g. `[A -> C, A -> B -> C]`)?

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+        for cycle in nx.simple_cycles(subgraph, length_bound=max_length):
+            if origin not in cycle:
+                continue
+            # Rotate the directed cycle so it starts at origin, then close it.
+            start = cycle.index(origin)
+            rotated = cycle[start:] + cycle[:start] + [origin]
+            paths.append(tuple(rotated))
+        return sorted(paths, key=lambda p: (len(p), p))
+
+    paths: set[tuple[str, ...]] = set()
```
kaitj: `visited` may be a better name for this variable to denote a difference from the paths we want to pass through

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+
+
+def load_metric(metric_file: str | Path) -> np.ndarray:
+    """Load a GIFTI metric as a one-dimensional floating-point array.
+
+    The first data array is returned as a NumPy array for comparison and
+    resampling during cycle evaluation.
+    """
+    return np.asarray(
+        nib.load(metric_file).darrays[0].data,
```
kaitj: More a general question after having seen how some of the giftis are created - is there anyway to confirm the file we are loading is a metric file vs a surface? I think the vertices would be loaded if a surface file was passed accidently?

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+            "Round-tripped metric did not return to the origin mesh: "
+            f"{roundtrip.shape} vs {original.shape}."
+        )
+    finite_mask = np.isfinite(original) & np.isfinite(roundtrip)
+    if np.any(finite_mask):
+        max_abs_diff = float(
+            np.max(np.abs(original[finite_mask] - roundtrip[finite_mask]))
+        )
+    else:
+        warnings.warn(
```
kaitj: I think we want to use the logger warning here rather than the [`warnings`](https://docs.python.org/3/library/warnings.html) library (which seems more intended for developers than warnings during runtime).

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+            "Round-trip metric contains no finite values; "
+            "max absolute difference is undefined.",
+            RuntimeWarning,
+            stacklevel=2,
+        )
+        max_abs_diff = float("nan")
```
kaitj: Can use `np.nan` here

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+
+    finite_count = np.count_nonzero(finite_mask)
+    vectors_equal = bool(np.allclose(original, roundtrip, equal_nan=True))
+
+    if finite_count < 2:
+        warnings.warn(
```
kaitj: logger.warning here too

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+            RuntimeWarning,
+            stacklevel=2,
+        )
+        pearson_r = 1.0 if finite_count > 0 and vectors_equal else 0.0
+        return pearson_r, max_abs_diff
+
+    original_finite = original[finite_mask]
+    roundtrip_finite = roundtrip[finite_mask]
+    original_std = float(np.std(original_finite))
+    roundtrip_std = float(np.std(roundtrip_finite))
```
kaitj: check if casting like this is necessary or if `dtype` can be set for `np.std`

tests/cycle.py
```diff
@@ -0,0 +1,481 @@
...
+    if np.isclose(original_std, 0.0) or np.isclose(roundtrip_std, 0.0):
+        pearson_r = 1.0 if vectors_equal else 0.0
+    else:
+        pearson_r = float(np.corrcoef(original_finite, roundtrip_finite)[0, 1])
```
kaitj: Similar casting comment here.

kaitj (changes requested):
Looking very close. A few nits and a few things that are probably worth addressing now rather than having another PR down the road to fix.

tests/cycle.py
tamsinrogers: ah yes, I should have removed this as it was mainly for testing purposes/manually constraining the amount of possible cycles being returned.

`max_length` refers to the number of edges
`max_paths` refers to the number of cycles that exist in the graph

I really don't think this functionality is necessary at this point, so I've removed both

kaitj (approved):
lgtm 🚀 

one minor change would be the redundant naming of `test_cycle_unit`, but content wise, looks good

tests/cycle.py
```diff
@@ -82,150 +78,53 @@ def label(self) -> str:
...
-    if allow_revisits and max_length is None:
-        raise ValueError("max_length is required when allow_revisits=True")
-
     if origin not in subgraph:
         raise ValueError(
             f"Origin space '{origin}' is not in the '{edge_type}' layer. "
             f"Available: {sorted(subgraph.nodes)}"
         )
 
-    if not allow_revisits:
```
kaitj: Not sure if this removal was intentional? I think without it, revisits are never allowed - not sure if this would be a problem?

tests/cycle.py
```diff
@@ -246,14 +145,6 @@ def load_metric(metric_file: str | Path) -> np.ndarray:
 
     data_array = img.darrays[0]
 
-    allowed_intents = {
```
kaitj: Think this is fine to remove, I don't think there is actually a way to validate the file we are loading is a metric from the intent codes.

kaitj (commented):
Noted the removal of the revisiting, not sure if intentional. Otherwise, looks good to me.

tests/unit/graph/test_cycle.py
```diff
@@ -0,0 +1,522 @@
...
+) -> SimpleNamespace:
+    """Mock Workbench metric resampling for isolated cycle tests.
+
+    Loads the input metric, applies the numpy-based resampling approximation
+    between synthetic sphere geometries, writes the resulting GIFTI metric,
+    and returns an object matching the interface expected from Workbench.
+    """
+    # required by workbench but not needed here
+    del method
+    del area_surfs
```
kaitj, line 247: Given these are tests, I think its fine to leave them as is without deleting them. `ruff` may complain that it isn't used, but we can add an ignore it noting the workbench requirement

kaitj (approved):
Looks good to me - thanks for adding the code back in. Had one comment about the deleted arguments, but it's a minor thing.

kaitj (approved):

Issue #211, opened 2026-07-23
Task: [TESTING: Regression]


tamsinrogers:
### Regression Tests
GOAL: add end-to-end validation tests to detect future changes that impact transformation accuracy.
- Implement full transformation cycle tests:
  - Start with spaces containing available RheMAP surface transforms.
  - Traverse complete transformation cycles through the graph.
  - Return to the original template space and evaluate preservation.
- Track quantitative validation metrics:
  - Correlation between original and transformed data.
  - Transformation error measures.
  - Other domain-specific metrics (e.g., eigenmode preservation).
- Establish performance benchmarks to prevent regressions as new templates and transformations are added.

### Tasks

- [ ] full cyclic test. start with RheMAP surf transforms - full cyclic test based on `regression_test_cycle.py`

tamsinrogers:
closed by #220

PR #220, opened 2026-07-27
Regression Test (refactor)

tamsinrogers:
## This PR:
addresses https://github.com/childmindresearch/neuromaps-prime/milestone/5, refactors #198 

Cycle regression test on the real NeuromapsPrime graph. Adds end-to-end cycle regression testing on the real Neuromaps graph to validate transform roundtrip quality across multi-hop paths.

Each run may contain:

* CSV summaries
* TXT summaries
* per-path transformation manifests
* intermediate metric files
* surface visualizations
* optional wb_command/niwrap/styx logs

Area surfaces are attempted in this order:

1. midthickness
2. pial
3. white

## Type of Change
- [ ] Bug fix
- [x] Enhancement
- [ ] Resource contribution
- [ ] Documentation
- [ ] Other

## Related Issue(s)
- #211 by @tamsinrogers

**run test:**
```
uv run pytest /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/test_cycle.py -v -x -s -o log_cli=true -o log_cli_level=DEBUG --log-disable=matplotlib --log-disable=matplotlib.font_manager --log-disable=matplotlib.colorbar --log-disable=styx_docker_runner --log-disable=urllib3 2> >(grep -v "The requested image's platform")
```
**current output (example: CIVETNMT, right hemisphere):**
```
INFO     test_cycle:test_cycle.py:523 CYCLE PROGRESS: 53/53 (100.0%) | origin=CIVETNMT | hemisphere=right | path=CIVETNMT -> Yerkes19 -> fsLR -> Yerkes19 -> CIVETNMT
INFO     tests.cycle:cycle.py:531 Executing hop 0: CIVETNMT -> Yerkes19 (right)
INFO     tests.cycle:cycle.py:554 Completed hop 0: CIVETNMT -> Yerkes19 using midthickness
INFO     tests.cycle:cycle.py:531 Executing hop 1: Yerkes19 -> fsLR (right)
INFO     tests.cycle:cycle.py:554 Completed hop 1: Yerkes19 -> fsLR using midthickness
INFO     tests.cycle:cycle.py:531 Executing hop 2: fsLR -> Yerkes19 (right)
INFO     tests.cycle:cycle.py:554 Completed hop 2: fsLR -> Yerkes19 using midthickness
INFO     tests.cycle:cycle.py:531 Executing hop 3: Yerkes19 -> CIVETNMT (right)
INFO     tests.cycle:cycle.py:554 Completed hop 3: Yerkes19 -> CIVETNMT using midthickness
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node00_CIVETNMT_midthickness.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node00_CIVETNMT_sphere.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node01_Yerkes19_midthickness.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node01_Yerkes19_sphere.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node02_fsLR_midthickness.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node02_fsLR_sphere.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node03_Yerkes19_midthickness.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node03_Yerkes19_sphere.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node04_CIVETNMT_midthickness.png
INFO     test_cycle:test_cycle.py:399 Saved surface plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots/a4745801b4c8_node04_CIVETNMT_sphere.png
INFO     test_cycle:test_cycle.py:950 cycle CIVETNMT -> Yerkes19 -> fsLR -> Yerkes19 -> CIVETNMT: r=0.140136 max|delta|=3.125e-01
INFO     test_cycle:test_cycle.py:838 Saved CSV: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right.csv
INFO     test_cycle:test_cycle.py:839 Saved TXT summary: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right.txt
INFO     test_cycle:test_cycle.py:840 Saved path results: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_right_plots
INFO     test_cycle:test_cycle.py:844 Saved command log: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/work_CIVETNMT_right/wb_commands.log
INFO     test_cycle:test_cycle.py:719 Saved combined cycle summary plot: /Users/tamsin.rogers/Desktop/github/neuromaps-prime/tests/regression/cycle_outputs_20260818_162439/cycle_CIVETNMT_summary.png
PASSED

============================================= 1 passed in 889.46s (0:14:49) =============================================
(base) CMI-RSCH-MBP118:neuromaps-prime tamsin.rogers$ 
```

image


tamsinrogers:
@kaitj here's an idea - let me know what you think, happy to refactor as needed!

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1193 @@
+"""Cycle regression test on the real NeuromapsPrime graph.
```
kaitj:
```suggestion
"""Cycle regression test on the real Neuromaps-PRIME graph.
```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1193 @@
...
+* surface visualizations
+* optional wb_command/niwrap/styx logs
+
+Area surfaces are attempted in this order:
+
+1. midthickness
+2. pial
+3. white
+
+Run with::
```
kaitj:
```suggestion
Run with:
```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1193 @@
...
+OUTPUT_DIR = Path(__file__).resolve().parent / f"cycle_outputs_{_RUN_SUFFIX}"
+
+OUTPUT_DIR.mkdir(
+    parents=True,
+    exist_ok=True,
+)
+
+# Set to a specific space to test only that origin, or use "all" to test
+# every configured origin.
+ORIGIN: str = "all"
```
kaitj: Any reason why this wouldn't be the default and immutable? The comment makes it seem as if users can change this, but given this would be in the test, I would assume this is would always be the case (and also wouldn't change between tests)

Also, does "all" return a single value or does it a compute a value per cycle per node?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1193 @@
...
+
+HEMISPHERES = (
+    "left",
+    "right",
+)
+
+MAX_CYCLE_LENGTH = 4
+MAX_PATHS: int | None = None
```
kaitj: Should be immutable if constant - type hint makes it seem as though this can be updated as well. If can change, would set it as a default arg in the whatever methods call this.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+    ("Yerkes29", "right"): 0.5,
+    ("MBM", "left"): 0.5,
+    ("MBM", "right"): 0.5,
+    ("NenckiMonash", "left"): 0.5,
+    ("NenkiMonash", "right"): 0.5,
+    ("all", "left"): 0.5,
+    ("all", "right"): 0.5,
+}
+
+LOG_COMMANDS = True
```
kaitj: If we are logging, then the docstring for the module should be updated to reflect this (e.g. logs aren't optional as stated in the docstring).

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+    "left",
+    "right",
+)
+
+MAX_CYCLE_LENGTH = 4
+MAX_PATHS: int | None = None
+
+# Minimum acceptable mean Pearson correlation for each regression run.
+# Values should be updated deliberately.  Last updated 08/18 (TR).
+MIN_MEAN_PEARSON: dict[tuple[str, str], float] = {
```
kaitj: Instead of hardcoding an acceptable passing threshold, which we would have to manually update, we should just write out these values to a timestamped file that we can compare / plot via the CI. If we want something to denote "pass / fail" in the regression, we could then read in the previous runs file or assume -np.inf to see if current run is better or worse.

For now, it is probably sufficient to just write to file - what do you think would be better, a csv/tsv or a json file for this?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+    """Return the baseline for an individual origin/hemisphere."""
+    return MIN_MEAN_PEARSON[(origin, hemisphere)]
+
+
+# -------------------------------------------------------------------------
+# Command logging
+# -------------------------------------------------------------------------
+
+
+class CommandLogHandler(logging.Handler):
```
kaitj: Not entirely sure if this is necessary for the test, but can see how it is useful for debugging.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+# -------------------------------------------------------------------------
+# Surface helpers
+# -------------------------------------------------------------------------
+
+
+def _load_surface_coords(
+    surface_file: Path,
+) -> np.ndarray:
+    """Load surface coordinates as ``(n_vertices, 3)``."""
+    image = nib.load(str(surface_file))
```
kaitj: Unneeded casting to str here.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+            encoding="utf-8",
+        )
+
+
+# -------------------------------------------------------------------------
+# Surface helpers
+# -------------------------------------------------------------------------
+
+
+def _load_surface_coords(
```
kaitj: Thomas also added a generic loader as part of the stats module - take a look and see if that is useful to call here.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+            )
+
+    raise ValueError(f"No pointset coordinates found in {surface_file}.")
+
+
+def _load_surface_topology(
+    surface_file: Path,
+) -> np.ndarray:
+    """Load triangle topology from a surface GIFTI."""
+    image = nib.load(str(surface_file))
```
kaitj: casting, and again take a look to see if the new image loader mentioned above would help.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+            data.ndim == 2
+            and data.shape[1] == 3
+            and np.issubdtype(
+                data.dtype,
+                np.integer,
+            )
+        ):
+            return data.astype(np.int32)
+
+    raise ValueError(f"No triangle topology found in {surface_file}.")
```
kaitj: Could you walk me through this one? From what I gather, this tries to verify that there are "triangles" in the surface file? Is there an easier way to do this? This for loop seems overly complex for that. As long as the data isn't an annotation, I would assume that both the coordinates and the topology exists.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+
+
+def _extract_surface_mesh(
```
kaitj: Follows from above comments, I'm not entirely sure if this is necessary to extract a mesh above loading with nibabel.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+    )
+
+    metric_file.parent.mkdir(
+        parents=True,
+        exist_ok=True,
+    )
+
+    nib.save(
+        image,
+        str(metric_file),
```
kaitj: casting

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+        hemisphere=hemisphere,
+        resource_type="sphere",
+    )
+
+    if sphere is None:
+        raise FileNotFoundError(
+            f"No sphere atlas for {origin} at {density} ({hemisphere})."
+        )
+
+    coords = _load_surface_coords(Path(sphere.fetch()))
```
kaitj: Can't recall at the moment, but I think `.fetch()` already returns a Path, so may not need to cast here.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+
+
+def _load_metric_values(
+    metric_file: Path,
+) -> np.ndarray:
+    """Load scalar values from a metric GIFTI."""
+    return np.asarray(
+        nib.load(str(metric_file)).darrays[0].data,
+        dtype=np.float64,
+    )
```
kaitj: Take a look at the loading method noted above. If this is necessary, could you walk me through the casting to the array versus using nibabel's built-in `get_*` methods?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+    """Find a surface whose vertex count matches the metric."""
+    atlases = graph.utils.cache.get_surface_atlases(
+        space=space,
+        hemisphere=hemisphere,
+        resource_type=resource_type,
+    )
+
+    for atlas in atlases:
+        try:
+            coords = _load_surface_coords(Path(atlas.fetch()))
```
kaitj: casting

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+        except (
+            ValueError,
+            FileNotFoundError,
+            OSError,
+        ):
+            continue
```
kaitj, line 386: Should probably throw a warning at minimum so that this isn't a silent skip

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+
+        if coords.shape[0] == n_vertices:
+            return atlas, atlas.density
```
kaitj: Could we not just call the attribute directly if we are already returning the object here?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+            space,
+            hemisphere,
+            metric_values.shape[0],
+        )
+        return
+
+    surface_atlas, density = match
+
+    try:
+        coords, faces = _extract_surface_mesh(Path(surface_atlas.fetch()))
```
kaitj: casting

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+        f"{path_token}_node{hop_index:02d}_{space}_{resource_type}.png"
+    )
+
+    try:
+        plot_surf(
+            coords,
+            faces,
+            metric_values,
+            rotate=[270, 0],
+            filename=str(output_path),
```
kaitj: Is the casting necessary here?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+            vmin, vmax = 0.0, 1.0
+
+        _plot_single_surface(
+            graph=graph,
+            space=space,
+            metric_values=metric_values,
+            hemisphere=hemisphere,
+            resource_type="midthickness",
+            vmin=float(vmin),
+            vmax=float(vmax),
```
kaitj: Given the conditional above, I think these are already guaranteed to be a floating number

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+            plot_dir=plot_dir,
+            path_token=path_token,
+            path_label=path_label,
+            hop_index=hop_index,
+            pearson_r=pearson_r,
+        )
+
+        _plot_single_surface(
```
kaitj: Any reason plot against both the midthickness and sphere? If needed, can also do that in a for loop since every other argument is the same. 

Another minor thing could consider is plotting as subplots so that all the same metrics can be saved as a single figure rather than multiple individual figures.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+    )
+
+    return metric_file
+
+
+def _make_xyz_product_metric(
+    graph: NeuromapsGraph,
+    origin: str,
+    density: str,
+    hemisphere: str,
```
kaitj: Type hint for hemisphere should be `Literal["left", "right"], which I think you had previously defined in `cycle.py`. Can import and make use of it.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+        hemisphere,
+        " -> ".join(path),
+    )
+
+
+def test_cycle_roundtrip() -> None:
+    """Round-trip synthetic metrics through real transformation cycles."""
+    graph = NeuromapsGraph()
+
+    origins = sorted(graph.nodes) if ORIGIN == "all" else [ORIGIN]
```
kaitj: As noted above, given this is a test, origin shouldn't be conditional - would be nice to have the option in a script otherwise.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+    origin: str,
+) -> None:
+    """Create one two-panel cycle summary plot for an origin space.
+
+    The left panel shows left-hemisphere cycle correlations and the right
+    panel shows right-hemisphere cycle correlations.
+    """
+    frames: dict[str, pd.DataFrame] = {}
+
+    for hemisphere in ("left", "right"):
```
kaitj:
```suggestion
    for hemisphere in HEMISPHERES:
```

You can re-use the definition from above.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+
+    paths = sorted(
+        {path for frame in frames.values() for path in frame["path"]},
+        key=lambda path: (
+            len(str(path).split(" -> ")) - 1,
+            str(path),
+        ),
+    )
+
+    paths = list(reversed(paths))
```
kaitj: Feel lie these can be combined and made simpler. I think it should also already return a `list` or does it return a `set` because that is the input provided?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+
+        logger.info(
+            "ALL SPACES (%s): %d cycles, mean Pearson r = %.6f, threshold = %.6f",
+            hemisphere,
+            len(combined),
+            mean_r,
+            all_min_mean_pearson,
+        )
+
+        assert mean_r >= all_min_mean_pearson - allowed_regression, (
```
kaitj: Is there a `pytest` method we can use for this assertion?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+                hemisphere,
+            )
+            continue
+
+        combined = pd.concat(
+            frames,
+            ignore_index=True,
+        )
+
+        mean_r = float(combined["pearson_r"].mean())
```
kaitj: Callling `.mean()` should already return a float, I believe.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+        )
+
+    if min_mean_pearson is None:
+        pytest.fail(
+            f"No baseline threshold configured for "
+            f"origin={origin}, hemisphere={hemisphere}. "
+            f"Observed mean r={mean_r:.6f}."
+        )
+
+    allowed_regression = 0.0001  # change to whatever value makes sense here
```
kaitj: Seeing this pop up again - could just make this a module constant (`ALLOWED_REGRESSION = 1e-4`)

tests/cycle.py
kaitj, line 1: At a glance, there seems to be new methods, along with some shuffling of the conditionals. Any particular reason for shuffling and if it is necessary, does it change the behavior of the unit tests, where some of this was used?

tests/regression/test_cycle.py
kaitj, line 1: There are a lot of utility methods written mixed with the testing that convolutes this file. If possible, would see if there is any redundancy with the existing `cycle.py` that can be imported and used, and if it is specific to the regression tests, add a new utility helper under `tests/regression`.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+    ("CIVET", "left"): 0.5,
+    ("CIVET", "right"): 0.5,
+    ("NCBR", "left"): 0.998746,
+    ("NCBR", "right"): 0.998746,
+    ("Yerkes29", "left"): 0.5,
+    ("Yerkes29", "right"): 0.5,
+    ("MBM", "left"): 0.5,
+    ("MBM", "right"): 0.5,
+    ("NenckiMonash", "left"): 0.5,
+    ("NenkiMonash", "right"): 0.5,
```
kaitj: Spelling

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+
+LOG_COMMANDS = True
+
+
+def _get_min_mean_pearson(
+    origin: str,
+    hemisphere: str,
+) -> float:
+    """Return the baseline for an individual origin/hemisphere."""
+    return MIN_MEAN_PEARSON[(origin, hemisphere)]
```
kaitj: If there are any nodes that are in the graph but missing in the mapping, this will throw a `KeyError`

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+
+    metric_file = output_dir / f"metric_{origin}_{density}_{hemisphere}.func.gii"
+
+    return _write_metric(
+        metric_file,
+        values,
+    )
+
+
+def _load_metric_values(
```
kaitj: Functionally, I think this is identical to the `load_metric` method previously created in the helpers.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+            total_usable_paths += result
+
+    plot_run_summaries(
+        run_dir=OUTPUT_DIR,
+    )
+
+    if ORIGIN == "all":
+        _save_all_summary(OUTPUT_DIR)
+
+    assert total_usable_paths > 0, (
```
kaitj, line 252: If there are no usable paths, I think this will raise a `FileNotFoundError` before ever reaching this assertion due to how `plot_run_summaries` is defined / called.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+logger = logging.getLogger(__name__)
+
+
+# -------------------------------------------------------------------------
+# Run configuration
+# -------------------------------------------------------------------------
+
+_RUN_SUFFIX = datetime.now().strftime("%Y%m%d_%H%M")
+
+OUTPUT_DIR = Path(__file__).resolve().parent / f"cycle_outputs_{_RUN_SUFFIX}"
```
kaitj, line 64: Minor, but does this need to be added to the `.gitignore` or is it already there? With the regression tests, we may end up wanting to commit at the end of the CI run to keep a record, just need to figure out what that looks like, but we shouldn't commit every output.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1199 @@
...
+            "Saved command log: %s",
+            command_handler.log_file,
+        )
+
+    if min_mean_pearson is None:
+        pytest.fail(
+            f"No baseline threshold configured for "
+            f"origin={origin}, hemisphere={hemisphere}. "
+            f"Observed mean r={mean_r:.6f}."
+        )
```
kaitj: I don't think this can ever be reached in it's current form. Either a float is returned or an error is raised.

kaitj (changes requested):
I think the idea is largely correct, but a few changes with the implementation may be needed. 

1. What I couldn't quite figure out is are we returning a single value or a value per cycle per node origin in the regression testing?
2. There is some redundancy with the helpers and existing methods. Take a look through and re-use where existing definitions where possible.

tests/regression/test_cycle.py
tamsinrogers: @kaitj good point, having values for individual nodes was helpful for testing purposes, but it does make more sense to only consider `"all"`. It returns:
```
INFO     test_cycle:test_cycle.py:784 ALL SPACES (left): 225 cycles, mean Pearson r = 0.564602, threshold = 0.500000
INFO     test_cycle:test_cycle.py:784 ALL SPACES (right): 225 cycles, mean Pearson r = 0.545228, threshold = 0.500000
```

tests/cycle.py
tamsinrogers: @kaitj shuffling mainly to get methods working across both types of tests - carried over some logic, etc that I started defining in `regression/test_cycle.py` and ultimately decided they worked better in `cycle.py`. I've been testing the unit tests with every change here and they are not affected.

tests/regression/test_cycle.py
tamsinrogers: I was able to hit this if I didn't define a threshold for a given space, but now that the test just uses `all`, I'll remove it.

tests/regression/test_cycle.py
tamsinrogers: yes, the list cast was unnecessary here

tests/regression/test_cycle.py
tamsinrogers: hm, I guess I really just need these two methods for plotting purposes. I've relaxed the constraints a bit and moved these helpers over to `images.py,` let me know what you think!

src/neuromaps_prime/analysis/images.py
```diff
@@ -43,6 +50,26 @@ class DataOutput(NamedTuple):
     image: nib.GiftiImage | NiftiImage | None = None
 
 
+def load_gifti(surface: str | Path) -> nib.GiftiImage:
```
kaitj: You could also call `load_data(..., return_image=True)` if you need the image instance.

src/neuromaps_prime/analysis/images.py
```diff
@@ -144,8 +171,26 @@ def relabel_gifti(
...
+    image = load_gifti(surface)
+    return np.asarray(image.darrays[0].data, dtype=np.float64)
+
+
+def load_surface_topology(
+    surface: str | Path,
+) -> np.ndarray:
+    """Load triangle topology from a surface GIFTI."""
+    image = load_gifti(surface)
+    return np.asarray(image.darrays[1].data, dtype=np.int32)
```
kaitj: For both of these functions, do we need to use the `np.asarray` cast? I don't recall at the moment, but think the `image.darrays[#].data` may already be a numpy array. If so, you could just call something like `image.darrays[#].data.astype(np.type)`

Alternatively, if these are only called a few times, just call the load and extract the data directly wherever these methods are called.

tests/regression/cycle_baseline_20260821_0135.csv
kaitj, line 1: Looks great. A few minor changes with this file:

1. I think with the filename, we can just call it `cycle_<data>_<time?>.csv`. We would always compare with the last one in the repo anyways for the CI
2. For provenance, we should also include ~the cycle it took in the contents and maybe~ the species
3. Not entirely sure we need to include "all" (last 2 rows), rather just keep it by node.
4. More of a Q, are there multiple cycles that it is taking and then averaging? Just trying to understand why the current last column is `mean_pearson_r` vs `pearson_r`

tests/regression/test_cycle.py
kaitj, line 1: I didn't take too close of a look at this file - it looked similar as to the last round, but will take a closer look later on after this round of reviews.

One thing of note is that there are still a lot of private helpers in the module. I think it's fine if there is a good rationale for it, otherwise would only keep anything directly related to the testing here (e.g. computations, etc.), and also not part of visualization.

tests/cycle.py
```diff
@@ -144,249 +185,456 @@ def find_return_paths(
...
-    """Load a GIFTI metric as a one-dimensional floating-point array.
+# -------------------------------------------------------------------------
+# Metrics
+# -------------------------------------------------------------------------
 
-    Raises:
-        ValueError: If the file is not a scalar metric GIFTI.
-    """
+
+def load_metric(
```
kaitj, line 279: I think this may also be able to be replaced by `load_data`. The only validation we wouldn't have is that it is 1D, but that could be checked on data load

tests/cycle.py
```diff
@@ -144,249 +185,456 @@ def find_return_paths(
...
+
+    result_path = getattr(result, "path", None)
+
+    if result_path is None:
+        return None
+
+    if isinstance(result_path, (str, Path)):
+        return Path(result_path)
+
+    return None
```
kaitj: This could be replaced by a two-line conditional where it is called (below). The `TransformResult.path` should always return a `Path` or `None` - I don't believe it can ever be a string.

tests/cycle.py
```diff
@@ -144,249 +185,456 @@ def find_return_paths(
...
+
+    candidate = workdir / result_path.name
+
+    if candidate.exists():
+        return candidate
+
+    raise FileNotFoundError(
+        "Surface transformation did not produce an accessible output. "
+        f"Expected '{expected}', got result={result!r}."
+    )
```
kaitj: I'm not sure I understand why this is needed. The NiWrap call to the workbench never creates a path that is inaccessible to the host. If the `path` is None, it should be because the transformation itself failed (if it doesn't raise an error before that).

Seeing how it is used below, I would also just replace the call with a two-line conditional to check if the output file path exists unless you have encountered an instance where the file path does exist, but couldn't be found?

kaitj (commented):
Getting closer 🚀 . I think there are still some redundancy in the methods introduced that are already part of other modules in the library and can be imported. I am not sure we want define a "FAIL" threshold versus just us logging what the values are. Albeit slightly different cases, for something like bids2table, we just report our metrics via the CI and let the user / maintainer determine if it is acceptable (see [here](https://github.com/childmindresearch/bids2table/pull/114#issuecomment-5361354441)).

Just a thought, but if you run the test twice, do you always get the same results, barring some very slight deviation? I think similar to the example in the link above (though don't worry about this aspect of it yet), we probably want to have some sort of quantitative comparison that can be presented per PR where we can then flag any notable changes for users to inspect more deeply.

> Note that the current version of this test uses ALLOWED_REGRESSION: Final = 1e-4 to account for small decimal changes in calculation, but I'd like to hear your thoughts @kaitj on what type of "wiggle-room" value would actually make sense here and why.

Assuming the metric that is created is always guaranteed to be the same across runs (and across machines), this value is probably fine if we go down the path of deciding what is a pass vs fail. There may be some deviation as a result of any numerical stability, but it shouldn't be a large deviation.

> Run-level "all" baseline comparison plot

I do like these plots, but I would make a minor change to them. Rather than have the node on the y-axis (this could get quite long), plot the correlation on the y-axis, the timestamp on the x-axis. Instead have a different symbol represent each node with filled vs hollow differentiating left versus right. We likely would have to limit these cyclic tests within species anyways, so if it helps just use the NHP nodes for now.

If I missed any other Qs or comments, feel free to let me know.

tests/regression/test_cycle.py
tamsinrogers: makes sense, I moved the cycle helpers to `cycle.py`, so everything in here should be related to regression computation and plotting

src/neuromaps_prime/analysis/images.py
tamsinrogers: yes, looks like `image.darrays[#].data.astype(np.type)` can be used

tests/regression/cycle_baseline_20260821_0135.csv
tamsinrogers: for (4), the average is taken from all executable cycles across all origins, so `pearson_r` is the Pearson correlation for one individual cycle/path (like Yerkes19), and `mean_pearson_r` is the Pearson correctlation averaged across all cycles.

I thought it could be useful to still record the pearson value per origin node, it might be interesting to look at (ie plotting)

kaitj: Not quite sure I am following in the context of the file. Maybe we are asking / answering 2 different things. 

In the file, the last column is `mean_pearson_r`. Does this mean, for example "Yerkes19", there are multiple executable cycles and we average across all of them or is it actually just 1 cyclic execution, just placed under a `mean_pearson_r` column? I guess a toy example of what I am trying to ask is are we looking at `avg([A->B->C->A, A->C->B->A])` or are we just looking at `A->B->C->A`.

tamsinrogers: yes, that value would be the average of r for all executable cycles across Yerkes19, so we are looking at avg([A→B→C→A, A→C→B→A])

tamsinrogers: but again, if that's not the intention of this test, I can refactor! it's what made sense to me at the time

kaitj: How much does this value change if you were to run the test twice for a single origin / hemi (maybe you answered this and I missed it)? If it doesn't change then I think this is fine.

tamsinrogers: it doesn't change at all, if I understand the question correctly. When I was hardcoding the values to test regression of each origin node, they were always the same.

tamsinrogers: (the "all" average values didn't change either). But of course there are certain origin nodes that have less regression in their cycles than others, which is why I think it's interesting to include them here.

kaitj: > it doesn't change at all, if I understand the question correctly

Cool 👍. This is what I would expect currently since we haven't changed anything in the codebase and everything should be deterministic.

> But of course there are certain origin nodes that have less regression in their cycles than others

Not quite following what you mean? The comparison we are trying to set up for is more to test regressions across versions in the future rather than across different cycles. We do want the values for each origin node - averaged across all cyclic permutations is fine. From the dev side of things any change affecting one cycle should similarly affect other cycles.

tamsinrogers: @kaitj oops sorry - what I mean here (I worded it wrong) is just that some cycles through the graph are better depending on which node the cycle originates at. For example, 
```
origin,hemisphere,mean_pearson_r
CIVETNMT,left,0.5016736728413902
CIVETNMT,right,0.5171821464198864
D99,left,0.6102685641174721
D99,right,0.5671449348092419
```
`D99` has slightly better preservation than `CIVETNMT`, which we know. The values themselves stay the same each time the test is run, so the only reason for actual _regression_ to occur would be if something changes from the dev side, like to said. I think we're on the same page!

kaitj: Cool, then I think this is close, would just add the species column.

tests/regression/utils.py
```diff
@@ -0,0 +1,392 @@
...
+    """Create a deterministic synthetic metric from sphere coordinates."""
+    sphere = graph.fetch_surface_atlas(
+        space=origin,
+        density=density,
+        hemisphere=hemisphere,
+        resource_type="sphere",
+    )
+
+    if sphere is None:
+        raise FileNotFoundError(
```
kaitj: this should actually be a `ValueError` I think since `sphere` is an object, while `sphere.file_path` is the file we are looking for.

tests/regression/utils.py
```diff
@@ -0,0 +1,392 @@
...
+    """Find a surface whose vertex count matches the metric."""
+    atlases = graph.utils.cache.get_surface_atlases(
+        space=space,
+        hemisphere=hemisphere,
+        resource_type=resource_type,
+    )
+
+    for atlas in atlases:
+        try:
+            coords = load_surface_coords(Path(atlas.fetch()))
```
kaitj:
```suggestion
            coords = load_surface_coords((atlas.fetch()).file_path)
```

This is probably more correct, rather than casting the object to a `Path`. `atlas` here should be an object, calling `.fetch()` would download the remote file if it doesn't already exist, and then we need to point it to the file path.

tests/regression/utils.py
```diff
@@ -0,0 +1,392 @@
...
+        encoding="utf-8",
+    )
+
+
+def load_latest_cycle_baseline(
+    baseline_dir: Path,
+) -> dict[tuple[str, str], float]:
+    """Load the most recent timestamped cycle baseline CSV."""
+    baseline_files = sorted(
+        baseline_dir.glob("cycle_????????_????.csv"),
```
kaitj: This regex pattern seems fragile - what happens if the csv is just `cycle_<data>.csv`, or if it doesn't match `????????`?

src/neuromaps_prime/analysis/images.py
```diff
@@ -144,8 +150,40 @@ def relabel_gifti(
...
+    """Load triangle topology from a surface GIFTI."""
+    _, image = load_data(
+        surface,
+        return_image=True,
+    )
+
+    if not isinstance(image, nib.GiftiImage):
+        raise ValueError(f"Expected a GIFTI surface: {surface}")
+
+    return image.darrays[1].data.astype(np.int32, copy=False)
```
kaitj: If these were always used individually, I would keep them as is, but seeing how they are called back-to-back below, I would just replace them with `load_data` calls directly wherever these are called. The main reason for this is the need to call `load_data` multiple times for a single surface just to get the topology or the coords this way versus calling it once by just calling `load_data` once and using the image or the array attribute.

kaitj (commented):
Still a few things to address, but its pretty close.

tests/regression/utils.py
tamsinrogers: makes sense, I'm trying a version that looks for certain headers instead of being so strict with the naming

tests/regression/test_cycle.py
```diff
@@ -419,28 +527,132 @@ def plot_run_summaries(
...
+        if origin == "all" or origin not in origins:
+            continue
 
+        species_origins.setdefault(
+            species,
+            [],
+        ).append(origin)
+
+    return {
+        species: sorted(set(species_origins[species])) for species in species_origins
```
kaitj: Having trouble picturing this, how does this sort the values - is it just sorting the origins alphabetically for each species data?

tests/regression/test_cycle.py
kaitj, line 1: I think in general, we don't need to reference the values as "baseline". All of it will be associated with a date anyways and we can do with that how we will.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1297 @@
...
+    frame: pd.DataFrame,
+    origins: list[str],
+) -> list[str]:
+    """Return origins that have cycle summary data."""
+    available_origins = set(frame["origin"])
+
+    return [origin for origin in origins if origin in available_origins]
+
+
+def _get_marker_map(
```
kaitj, line 372: Given the limited number of markers (and maybe there is more?), wondering if it makes sense for visualization to split into multiple graphs (per species). Would we ever end up hitting more origins than markers? Alternatively, we plot by color with always the same type of marker, or denote the marker by species.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1297 @@
...
+
+        pearson_r, max_abs_diff = score_roundtrip(
+            metric_file,
+            roundtrip.final_metric,
+        )
+
+    except (
+        RuntimeError,
+        FileNotFoundError,
+        OSError,
```
kaitj: What would the `OSError` be associated with here?

tests/regression/cycle_baseline_20260821_0135.csv
kaitj, line 1: re: one of the comments in the other module about denoting variables / file names with "baseline". I think the date is probably enough here. Would also still want to include the species in this file as a column.

kaitj:
Also needs a rebase from `main`

tests/regression/test_cycle.py
tamsinrogers: This version has both `cycle_summary.png` (all species) and `cycle_summary_{SPECIES}.png` plots - I think both are useful visualizations.  We would only need a new marker if a new space is added - I can had a handful more here to be safe or come up with a different way of choosing markers.

tests/regression/test_cycle.py
tamsinrogers: yes that's correct. I reworked it a bit in the latest version, but this is mainly for readability of the csv. I think having it broken out by species in the combined summary version is useful.

src/neuromaps_prime/analysis/images.py
```diff
@@ -117,6 +118,9 @@ def relabel_gifti(
         A ``GiftiImage`` instance with consecutive label indices and an
         updated label table.
 
+    Raises:
+        ValueError: If loaded image is not a GiftiImage.
+
     Raises:
         ValueError: If loaded image is not a GiftiImage.
```
kaitj, line 121: Duplicated text in the docstring

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1307 @@
...
+
+# Always test every configured origin space.
+
+HEMISPHERES = (
+    "left",
+    "right",
+)
+
+MAX_CYCLE_LENGTH: Final = 4
+MAX_PATHS: Final[int | None] = None
```
kaitj: Is this suppose to be mutable? Currently defined as a constant.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1307 @@
...
+        combined = pd.concat(
+            frames,
+            ignore_index=True,
+        )
+
+        values[("all", hemisphere)] = float(
+            combined["pearson_r"].mean(),
+        )
+
+    return values
```
kaitj: Could you walk me through this? Could the two for loops be combined, it's cycling over the same values.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1307 @@
...
+    )
+
+    current_values = _collect_cycle_values(
+        run_dir=OUTPUT_DIR,
+        origins=origins,
+    )
+
+    current_all_values = _save_all_summary(
+        run_dir=OUTPUT_DIR,
+        r=csv,
```
kaitj: Just noticing this here, should we give this argument a more meaningful value?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1307 @@
...
+    save_cycle(
+        dir=dir,
+        values=current_values,
+        graph=graph,
+    )
+
+    plot_run_summaries(
+        run_dir=OUTPUT_DIR,
+        current_values=current_values,
+        r=csv,
```
kaitj: Same for this value.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1307 @@
...
+            frame = pd.read_csv(csv_path)
+
+            if frame.empty:
+                continue
+
+            rows.append(
+                {
+                    "origin": origin,
+                    "hemisphere": hemisphere,
+                    "pearson_r": float(frame["pearson_r"].mean()),
```
kaitj: is the casting necessary here?

kaitj (commented):
Some very minor things

tests/regression/test_cycle.py
tamsinrogers: good point, I've combined the logic and just used `load_latest_cycle_values` in the latest version in favor of having two separate helpers that both grab the csv/csv values in different ways.

tests/regression/test_cycle.py
tamsinrogers: These don't ever get reassigned, so I think they are ok to be constants.

kaitj: In that case, I would just set the method argument to `None` where it is called instead of defining this constant since it looks like it is only called once in this test suite. If it needs to be kept, then update the typehint since it would never `int` if it doesn't ever get reassigned.

```suggestion
```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1273 @@
...
+        values[("all", hemisphere)] = float(
+            combined["pearson_r"].mean(),
+        )
+
+    return values
+
+
+def test_cycle_roundtrip() -> None:
+    """Round-trip synthetic metrics through real transformation cycles."""
+    graph = NeuromapsGraph()
```
kaitj: Is there a fixture that already exists for the graph that can be passed here?

src/neuromaps_prime/analysis/images.py
```diff
@@ -106,6 +106,7 @@ def relabel_gifti(
     Loads the parcellation file, zeroes out any background labels found in
     the label table, then remaps the remaining indices to consecutive
     integers starting at ``1``. Returns a new ``GiftiImage`` with an
```
kaitj, line 108:
```suggestion
```

duplicated docstring

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1273 @@
...
+        return 0
+
+    original_metric = load_metric(metric_file)
+
+    paths = find_return_paths(
+        graph,
+        origin,
+        max_length=MAX_CYCLE_LENGTH,
+        allow_revisits=True,
+        max_paths=MAX_PATHS,
```
kaitj:
```suggestion
```

Default is already `None`, so can omit the parameter since it is never modified

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1273 @@
...
+            mean_r,
+            s=120,
+            marker=marker,
+            facecolors="none",
+            edgecolors=color,
+            linewidths=2.0,
+            label=f"{origin} (R)",
+            zorder=3,
+        )
+
```
kaitj:
```suggestion
    ax.scatter(
        x_position,
        mean_r,
        s=120,
        marker=marker,
        facecolors=color,
        edgecolors=color,
        linewidths=1.5,
        label=f"{origin} {hemisphere.upper()[0]}",
        zorder=3,
    )
```

May have gotten the syntax slightly wrong, but since hemisphere is already a string, can just transform to uppercase and take the first letter to match what you are trying to do here.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1273 @@
...
+        origin: color_cycle[index % len(color_cycle)]
+        for index, origin in enumerate(origins)
+    }
+
+
+def _plot_hemisphere_point(
+    ax: plt.Axes,
+    x_position: int,
+    origin: str,
+    hemisphere: str,
```
kaitj: wasn't there a `Hemisphere` typehint that matched `Literal` that can be used here?

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1273 @@
...
+    mean_r: float,
+    marker: str,
```
kaitj:
```suggestion
    marker: str | MarkerStyle,
```

I believe the `_get_marker_map` method ends up returning a MarkerStyle (according to the typehint). May need to import MarkerStyle

tests/cycle.py
```diff
@@ -36,103 +36,144 @@
...
+    """Output from executing one complete transformation cycle."""
+
+    path: tuple[str, ...]
+    final_metric: Path
+    hops: tuple[HopResult, ...]
+
+    @property
+    def intermediates(self) -> tuple[tuple[str, Path], ...]:
+        """Return destination-space metric files for every hop."""
+        return tuple((hop.target, hop.output_file) for hop in self.hops)
```
kaitj:
```suggestion
```

unused code since changes to the PR. Drop or wire in where needed

tests/regression/utils.py
```diff
@@ -0,0 +1,475 @@
...
+                str(hop.output_file),
+            ]
+        )
+        for hop in roundtrip.hops
+    )
+
+    output_file.write_text(
+        "\n".join(lines) + "\n",
+        encoding="utf-8",
+    )
```
kaitj:
```suggestion
```

Unused code since changes

tests/cycle.py
```diff
@@ -36,103 +36,144 @@
...
 
     path: tuple[str, ...]
     pearson_r: float
     max_abs_diff: float
 
     @property
     def label(self) -> str:
-        """Return a human-readable ``A -> B -> A`` label for the path."""
+        """Return a human-readable path label."""
         return " -> ".join(self.path)
```
kaitj:
```suggestion
```

This is from a prior PR, but this consumed anywhere in the tests.

tests/regression/utils.py
```diff
@@ -0,0 +1,475 @@
...
+    from neuromaps_prime.graph import NeuromapsGraph
+
+logger = logging.getLogger(__name__)
+
+# -------------------------------------------------------------------------
+# Metric generation
+# -------------------------------------------------------------------------
+
+
+def write_metric(
```
kaitj: A similar save method is duplicated in [`tests/unit/graph/test_cycle/_save_metric`](https://github.com/childmindresearch/neuromaps-prime/blob/1cd41f56d8355c716207f0d58a632e5ef958315f/tests/unit/graph/test_cycle.py#L144-L145). It would probably be better to consolidate the two by pulling it into `tests/cycle.py` where the helpers are and then import the call as needed.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1273 @@
...
+            f"No cycle CSV files found in {run_dir}.",
+        )
+
+    origins = []
+
+    for csv_file in csv_files:
+        name = csv_file.name
+
+        if not name.startswith("cycle_"):
+            continue
```
kaitj:
```suggestion
```

This is unnecessary since the `glob` above would already guarantee that the files looped through starts with "cycle_"

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1273 @@
...
+        completed,
+        total,
+        percentage,
+        origin,
+        hemisphere,
+        " -> ".join(path),
+    )
+
+
+def _collect_cycle_values(
```
kaitj: There is some I/O redundancy with this method, `_load_cycle_summary_data`, `_collect_summary_rows`, and `_calculate_all_values`, where the CSVs have to be re-read each time (up to 5 reads) There is also some repeated computation across some of the methods. 

Could collapse all four methods into two - a reader and a summarizer. Something like:

```python
from typing import NamedTuple

def _read_cycle_frames(
    run_dir: Path, origins: list[str]
) -> dict[tuple[str, str], pd.DataFrame]:
    """Read each ``cycle_*.csv`` once, keyed by (origin, hemisphere)."""
    frames: dict[tuple[str, str], pd.DataFrame] = {}
    for origin in origins:
        for hemisphere in HEMISPHERES:
            csv_path = run_dir / f"cycle_{origin}_{hemisphere}.csv"
            if not csv_path.exists():
                continue
            frame = pd.read_csv(csv_path)
            if not frame.empty:
                frames[(origin, hemisphere)] = frame
    return frames

class _CycleSummary(NamedTuple):
    """Aggregated cycle means and derived report rows."""

    values: dict[tuple[str, str], float] # (origin|all, hemi) -> mean r
    origin_rows: list[dict[str, object]] # per-origin rows for cycle_summary.csv
    all_rows: list[dict[str, object]]    # all-space rows for cycle_summary.csv
    summaries: list[str]                 # all-space summary text for cycle_all_summary.txt


def _summarize(
    frames: dict[tuple[str, str], pd.DataFrame],
    species_by_origin: dict[str, str],
    previous: dict[tuple[str, str], float],
) -> _CycleSummary:
    values = {}
    origin_rows = []
    by_hemi = {h: [] for h in HEMISPHERES}

    for (origin, hemisphere), frame in sorted(frames.items()):
        values[(origin, hemisphere)] = frame["pearson_r"].mean()
        origin_rows.append(
            {
                "origin": origin,
                "species": species_by_origin.get(origin, "all"),
                "hemisphere": hemisphere,
                "mean_pearson_r": values[(origin, hemisphere)],
            }
        )
        by_hemi[hemisphere].append(frame)

    all_rows = []
    summaries = []
    for hemisphere in HEMISPHERES:
        if not by_hemi[hemisphere]:
            continue
        mean_r = pd.concat(by_hemi[hemisphere], ignore_index=True)["pearson_r"].mean()
        values[("all", hemisphere)] = mean_r
        all_rows.append({"origin": "all", "hemisphere": hemisphere, "mean_pearson_r": mean_r})

        previous_r = previous[("all", hemisphere)]
        difference = mean_r - previous_r
        # Note assertion is removed here since we only care about saving the values
        summaries.append(
            f"All spaces ({hemisphere}):\n"
            f"  Total executable cycles: {sum(len(f) for f in by_hemi[hemisphere])}\n"
            f"  Mean Pearson r: {mean_r:.6f}\n"
            f"  Previous Pearson r: {previous_r:.6f}\n"
            f"  Difference: {difference:+.6f}\n"
        )
    return _CycleSummary(values, origin_rows, all_rows, summaries)

    def test_cycle_roundtrip() -> None:
        dir = Path(__file__).resolve().parent
        previous, input_csv = load_latest_cycle_values(dir=dir)

        origins = sorted(graph.nodes)
        for origin in origins:
            for hemisphere in HEMISPHERES:
                total = _run_origin_hemisphere(graph, origin, hemisphere)
                assert total >= 0
        
        species_by_origin = _species_map(input_csv)
        frames = _read_cycle_frames(OUTPUT_DIR, origins)
        values, origin_rows, all_rows, summaries = _summarize(frames, species_by_origin, previous)

        _save_cycle_summary_csv(OUTPUT_DIR, origin_rows + all_rows) # replaces _collect_summary_rows
        _save_all_summary_txt(OUTPUT_DIR, summaries) # replaces _calculate_all_values
        save_cycle(dir=dir, values=values, graph=graph)
        plot_run_summaries(run_dir=OUTPUT_DIR, values=values, previous==previous, summary_rows=origin_rows) # Read from table instead of CSVs...for now
```

Could also make the class a frozen dataclass to match the others instead of a NamedTuple.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1108 @@
...
+
+        vertices = path.vertices
+
+        width = vertices[:, 0].max() - vertices[:, 0].min()
+        height = vertices[:, 1].max() - vertices[:, 1].min()
+        scale = max(width, height)
+
+        if scale:
+            vertices = vertices / scale
+
```
kaitj: This is computed, but never applied.

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1108 @@
...
+        "Saved cycle comparison plot: %s",
+        output_file,
+    )
+
+
+def _save_cycle_results(
+    origin: str,
+    hemisphere: Hemisphere,
+    rows: list[dict[str, object]],
+    plot_dir: Path,
```
kaitj: This param is only used in the logging, but plots are written to `OUTPUT_DIR`

tests/regression/test_cycle.py
tamsinrogers: hmm, there aren't currently any fixtures for the regression test in the way that the unit and integration tests have for the graph

tests/regression/utils.py
```diff
@@ -0,0 +1,413 @@
...
+                )
+
+            values[key] = float(row["mean_pearson_r"])
+
+        logger.info(
+            "Using cycle pearson r: %s",
+            file,
+        )
+
+        return values, file
```
kaitj:
```suggestion
def load_latest_cycle_values(dir: Path) -> dict[tuple[str, str], float]
    """Load the most recent valid cycle Pearson r CSV."""
    files = sorted(
        dir.glob("cycle_*.csv"),
        key=lambda path: path.stat().st_mtime,
        reverse=True,
    )

    required_columns = {
        "origin",
        "species",
        "hemisphere",
        "mean_pearson_r",
    }

    for file in files:
        try:
            frame = pd.read_csv(file)
        except (OSError, pd.errors.ParserError):
            continue

        if not required_columns.issubset(frame.columns):
            continue

        values: dict[tuple[str, str], float] = {}

        for _, row in frame.iterrows():
            key = (
                str(row["origin"]),
                str(row["hemisphere"]),
            )

            if key in values:
                raise ValueError(
                    f"Duplicate pearson r entry in {file}: {key}",
                )

            values[key] = float(row["mean_pearson_r"])

        logger.info(
            "Using cycle pearson r: %s",
            file,
        )

        return values
```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1108 @@
...
+        current_values=summary.pearson_r,
+        pearson_r=previous,
+        summary_rows=summary.origin_rows,
+    )
+
+    logger.info(
+        "NEW PEARSON R VALUES: left=%.6f, right=%.6f",
+        summary.pearson_r[("all", "left")],
+        summary.pearson_r[("all", "right")],
+    )
```
kaitj:
```suggestion
def test_cycle_roundtrip() -> None:
    """Round-trip synthetic metrics through real transformation cycles."""
    graph = NeuromapsGraph()
    dir = Path(__file__).resolve().parent

    previous = load_latest_cycle_values(dir=dir)

    origins = sorted(graph.nodes)
    total_usable_paths = 0

    for origin in origins:
        for hemisphere in HEMISPHERES:
            logger.info(
                "\n=== CYCLE ROUND-TRIP TEST: %s (%s) ===",
                origin,
                hemisphere,
            )

            total_usable_paths += _run_origin_hemisphere(
                graph=graph,
                origin=origin,
                hemisphere=hemisphere,
            )

    assert total_usable_paths > 0, (
        "No executable surface transformation cycles were found "
        "for any configured origin/hemisphere."
    )

    frames = _read_cycle_frames(
        run_dir=OUTPUT_DIR,
        origins=origins,
    )

    summary = _summarize(frames=frames, previous=previous)

    for hemisphere in HEMISPHERES:
        key = ("all", hemisphere)

        current_r = summary.pearson_r[key]
        previous_r = previous[key]

        assert current_r >= previous_r - 1e-4, (
            f"Average round-trip correlation regressed: "
            f"origin=all, hemisphere={hemisphere}, "
            f"mean r={current_r:.6f}, "
            f"previous={previous_r:.6f}, "
            f"threshold={previous_r - 1e-4:.6f}"
        )

    _save_cycle_summary_csv(
        run_dir=OUTPUT_DIR,
        rows=summary.origin_rows + summary.all_rows,
    )

    _save_all_summary_txt(
        run_dir=OUTPUT_DIR,
        summaries=summary.summaries,
    )

    save_cycle(
        dir=dir,
        values=summary.pearson_r,
        graph=graph,
    )

    plot_run_summaries(
        run_dir=OUTPUT_DIR,
        current_values=summary.pearson_r,
        pearson_r=previous,
        summary_rows=summary.origin_rows,
    )

    logger.info(
        "NEW PEARSON R VALUES: left=%.6f, right=%.6f",
        summary.pearson_r[("all", "left")],
        summary.pearson_r[("all", "right")],
    )
```

tests/regression/utils.py
```diff
@@ -0,0 +1,413 @@
...
+        output_file,
+        index=False,
+    )
+
+    logger.info(
+        "Saved cycle pearson r: %s",
+        output_file,
+    )
+
+    return output_file
```
kaitj:
```suggestion
```

tests/regression/utils.py
```diff
@@ -0,0 +1,413 @@
+"""Utilities for cycle regression tests."""
+
+from __future__ import annotations
+
+import logging
+from datetime import datetime
```
kaitj:
```suggestion
```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1108 @@
...
+    load_metric,
+    path_token,
+    roundtrip_metric,
+    score_roundtrip,
+)
+from tests.regression.utils import (
+    load_latest_cycle_values,
+    make_sphere,
+    plot_cycle_cortical_surfaces,
+    save_cycle,
```
kaitj:
```suggestion
```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1108 @@
...
+            f"  Difference: {difference:+.6f}\n"
+        )
+
+    return _CycleSummary(
+        pearson_r=pearson_r,
+        origin_rows=origin_rows,
+        all_rows=all_rows,
+        summaries=summaries,
+    )
+
```
kaitj, line 223:
```suggestion
def _summarize(
    frames: dict[tuple[str, str], pd.DataFrame],
    species_by_origin: dict[str, str],
    previous: dict[tuple[str, str], float],
) -> _CycleSummary:
    """Calculate cycle means and derived summary rows."""
    pearson_r: dict[tuple[str, str], float] = {}
    origin_rows: list[dict[str, object]] = []
    all_rows: list[dict[str, object]] = []
    summaries: list[str] = []

    by_hemi: dict[str, list[pd.DataFrame]] = {
        hemisphere: [] for hemisphere in HEMISPHERES
    }

    for (origin, hemisphere), frame in sorted(frames.items()):
        mean_r = float(frame["pearson_r"].mean())

        pearson_r[(origin, hemisphere)] = mean_r

        origin_rows.append(
            {
                "origin": origin,
                "species": species_by_origin.get(
                    origin,
                    "all",
                ),
                "hemisphere": hemisphere,
                "mean_pearson_r": mean_r,
            }
        )

        by_hemi[hemisphere].append(frame)

    for hemisphere in HEMISPHERES:
        frames_for_hemisphere = by_hemi[hemisphere]

        if not frames_for_hemisphere:
            continue

        combined = pd.concat(
            frames_for_hemisphere,
            ignore_index=True,
        )

        mean_r = float(combined["pearson_r"].mean())

        pearson_r[("all", hemisphere)] = mean_r

        all_rows.append(
            {
                "origin": "all",
                "species": "all",
                "hemisphere": hemisphere,
                "mean_pearson_r": mean_r,
            }
        )

        previous_r = previous[("all", hemisphere)]
        difference = mean_r - previous_r

        summaries.append(
            f"All spaces ({hemisphere}):\n"
            f"  Total executable cycles: {len(combined)}\n"
            f"  Mean Pearson r: {mean_r:.6f}\n"
            f"  Previous Pearson r: {previous_r:.6f}\n"
            f"  Difference: {difference:+.6f}\n"
        )

    return _CycleSummary(
        pearson_r=pearson_r,
        origin_rows=origin_rows,
        all_rows=all_rows,
        summaries=summaries,
    )

```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1108 @@
...
+    summary_frame.to_csv(
+        run_dir / "cycle_summary.csv",
+        index=False,
+    )
+
+    logger.info(
+        "Saved cycle summary CSV: %s",
+        run_dir / "cycle_summary.csv",
+    )
```
kaitj:
```suggestion
def _summary_frame(rows: list[dict[str, object]]) -> pd.DataFrame:
    """Build the canonical per-origin + all-space summary frame."""
    return (
        pd.DataFrame(
            rows,
            coluumns=["origin", "species", "hemisphere", "mean_pearson_r"],
        )
        .sort_values(["species", "origin", "hemisphere"], kind="stable")
        .reset_index(drop=True)
    )
```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1108 @@
...
+) -> dict[str, str]:
+    """Return the species associated with each origin."""
+    frame = pd.read_csv(input_csv)
+
+    return (
+        frame[["origin", "species"]]
+        .drop_duplicates("origin")
+        .set_index("origin")["species"]
+        .to_dict()
+    )
```
kaitj:
```suggestion
```

tests/regression/test_cycle.py
```diff
@@ -0,0 +1,1108 @@
...
+        pearson_r=previous,
+        summary_rows=summary.origin_rows,
+    )
+
+    logger.info(
+        "NEW PEARSON R VALUES: left=%.6f, right=%.6f",
+        summary.pearson_r[("all", "left")],
+        summary.pearson_r[("all", "right")],
+    )
+
```
kaitj:
```suggestion
def test_cycle_roundtrip() -> None:
    """Round-trip synthetic metrics through real transformation cycles."""
    graph = NeuromapsGraph()
    dir = Path(__file__).resolve().parent

    previous = load_latest_cycle_values(
        dir=dir,
    )

    origins = sorted(graph.nodes)
    total_usable_paths = 0

    for origin in origins:
        for hemisphere in HEMISPHERES:
            logger.info(
                "\n=== CYCLE ROUND-TRIP TEST: %s (%s) ===",
                origin,
                hemisphere,
            )

            total_usable_paths += _run_origin_hemisphere(
                graph=graph,
                origin=origin,
                hemisphere=hemisphere,
            )

    assert total_usable_paths > 0, (
        "No executable surface transformation cycles were found "
        "for any configured origin/hemisphere."
    )

    species_by_origin = {
        origin: graph.get_node_data(origin).species for origin in origins
    }

    frames = _read_cycle_frames(
        run_dir=OUTPUT_DIR,
        origins=origins,
    )

    summary = _summarize(
        frames=frames,
        species_by_origin=species_by_origin,
        previous=previous,
    )

    for hemisphere in HEMISPHERES:
        key = ("all", hemisphere)

        current_r = summary.pearson_r[key]
        previous_r = previous[key]

        assert current_r >= previous_r - 1e-4, (
            f"Average round-trip correlation regressed: "
            f"origin=all, hemisphere={hemisphere}, "
            f"mean r={current_r:.6f}, "
            f"previous={previous_r:.6f}, "
            f"threshold={previous_r - 1e-4:.6f}"
        )

    frame = _summary_frame(
        summary.origin_rows + summary.all_rows,
    )

    baseline_path = dir / f"cycle_{datetime.now():%Y%m%d_%H%M}.csv"
    summary_path = OUTPUT_DIR / "cycle_summary.csv"

    frame.to_csv(
        baseline_path,
        index=False,
    )

    frame.to_csv(
        summary_path,
        index=False,
    )

    logger.info(
        "Saved baseline CSV: %s",
        baseline_path,
    )

    logger.info(
        "Saved cycle summary CSV: %s",
        summary_path,
    )

    _save_all_summary_txt(
        run_dir=OUTPUT_DIR,
        summaries=summary.summaries,
    )

    plot_run_summaries(
        run_dir=OUTPUT_DIR,
        current_values=summary.pearson_r,
        pearson_r=previous,
        summary_rows=summary.origin_rows,
    )

    logger.info(
        "NEW PEARSON R VALUES: left=%.6f, right=%.6f",
        summary.pearson_r[("all", "left")],
        summary.pearson_r[("all", "right")],
    )

```

tests/regression/test_cycle.py
kaitj: Could add a fixture similar to the unit or integration tests if you think we would need to call NeuromapsGraph, otherwise this is fine!

tests/regression/cycle_20260824_2316.csv
kaitj, line 1: This was great to see what would be generated, I would remove this for now and we'll have the CI generate this either on every PR merge or on every version tag and have it commit the file directly so everything is automated.

PR #221, opened 2026-07-27
add citation for D99 168k surface

tamsinrogers:
## Type of Change
- [ ] Bug fix
- [ ] Enhancement
- [ ] Resource contribution
- [ ] Documentation
- [ ] Other

## Related Issue(s)
- #[issue number] by @[original issue author]


kaitj (approved):
Hmm, I see how this gets tricky since only the new node densities should use the zenodo reference (or both) and the old node densities should only use the original reference. We should think about how to make this clear, but that is for a downstream PR.

Issue #182, opened 2026-06-16
Task: all OSF files should be gzipped


tamsinrogers:
### Description

some of them are full files

### Tasks

- [ ] gzip large files and re-upload to OSF

Issue #181, opened 2026-06-15
Task: Integrate JunaChimp


tamsinrogers:
### Description

I added a node for `JunaChimp` in #180 , but the only current data is a T1w.  

### Tasks

- [ ] If available, locate `JunaChimp` surface files. Add to OSF and add pointers to `JunaChimp.yaml`
- [ ] Locate (or calculate) min 1 transform edge between `JunaChimp` and an existing node.

PR #236, opened 2026-08-11
OSF id updates

tamsinrogers:
## This PR:
Adds/updates OSF ids for:
- some files that have been zipped/re-uploaded/otherwise changed (MBM, NenckiMonash, NCBR)
- NCBR annotations 
- Rhemap-surf files (github links) - failing without the addition of `raw.githubusercontent`

## Type of Change
- [ ] Bug fix
- [ ] Enhancement
- [x] Resource contribution
- [ ] Documentation
- [ ] Other

## Related Issue(s)
- #181 
- #182 
- #29


kaitj (commented):
This looks good to me. I am guessing the NCBR data was moved to a different project hence the URL change? 

The only other comment (and just in general) is that it looks like the same "view-only" key is passed throughout for the OSF projects. If the project is now public, we should remove the view-only key. If the project is private, we should update the view-only key since that should be specific to the project.

tamsinrogers:
@kaitj the NCBR files were re-uploaded, hence the new ids.  They are in the same (public) project as the other resources (`nmxfh`).  

I've removed the view-only tag from the public resources, keeping it for the receptor maps stored in the private project.

kaitj (approved):

PR #237, opened 2026-08-12
Summarizing continuous data with a discrete parcellations

tfunck:
## What it does
Summarises a continuous map within each region of a discrete parcellation.
Works identically for surface and volumetric data: the only requirement is that
the parcellation has the same spatial shape as the map. No resampling is
performed — transform the map into the parcellation's space first.

```python
from neuromaps_prime.analysis import parcellate

result = parcellate("thickness.func.gii", "atlas.label.gii")
result.labels   # array([1, 2, 3, ...])  region IDs, background excluded
result.values   # array([2.31, 2.44, ...])  one value per region
```

Accepts arrays, file paths (GIFTI or NIfTI), or nibabel images for either
argument. Supports `mean` (default), `median`, `sum`, `std`, `min`, `max`, or
any callable taking `(array, axis=...)`. NaNs are ignored; regions that are
entirely NaN return NaN. `background` (default `0`) selects the label to drop;
`min_valid` sets how many finite elements a region needs before a value is
computed.

## Why not just expose `vertices_to_parcels`?

`vertices_to_parcels` already exists in
`analysis/surfaces/nulls/spins.py`, ported alongside the null models, but it is
surface-only, accepts file paths only (an in-memory label array raises
`TypeError`), is hard-coded to `nanmean`, and is not exported from any
`__init__.py`. This adds the general case while leaving that function untouched;
a test asserts the two agree on surface data.

## Notes

- Grouping is done with a single stable sort rather than a boolean mask per
  region. On a ~1M-voxel volume with 1000 regions that is ~330 ms versus ~2.7 s
  for the naive approach.
- 2-D input `(n_elements, n_features)` is supported and returns
  `(n_regions, n_features)`.
- Non-contiguous and unsorted label IDs are handled; `labels` comes back sorted
  ascending so `values` is always unambiguous.


src/neuromaps_prime/analysis/parcellate.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    if arr.ndim == 1:
+        return arr[:, None]
+    if arr.ndim == 2:
+        return arr
+    if arr.ndim == 3:
+        return arr.reshape(-1, 1)
+    if arr.ndim == 4:
+        return arr.reshape(-1, arr.shape[-1])
+    raise ValueError(f"Unsupported {what} with {arr.ndim} dimensions.")
```
kaitj:
```suggestion
def _collapse(arr: np.ndarray):
    if arr.ndim not in (1, 2, 3, 4):
        raise ValueError(f"Unsupported array with {arr.ndim} dimensions.")
    out = arr.reshape(-1, *arr.shape[2:])  # (N,) or (N, F)
    return out[:, None] if out.ndim == 1 else out
```

Just simplifying the conditional chain. Also potentiall rename since we aren't flattening, but collapsing the array.

src/neuromaps_prime/analysis/parcellate.py
```diff
@@ -0,0 +1,191 @@
...
+            raise FileNotFoundError(f"{what} file does not exist: {path}")
+        img = nib.load(str(path))
+    elif isinstance(obj, (nib.GiftiImage, nib.Nifti1Image, nib.Nifti2Image)):
+        img = obj
+    else:
+        return np.asarray(obj)
+
+    if isinstance(img, nib.GiftiImage):
+        return np.asarray(img.agg_data())
+    return np.asarray(img.get_fdata())
```
kaitj:
```suggestion
def load(obj: ArrayLike | str | Path) -> np.ndarray:
    if isinstance(obj, (str, Path)):
        if not Path(obj).exists():
            raise FileNotFoundError(f"File does not exist: {obj}")
        img = nib.load(obj)
    elif isinstance(obj, nib.GiftiImage, nib.Nifti1Image, nib.Nifti2Image)):
        img = obj
    else:
        return np.asarray(obj, dtype=float)
    
    if isinstance(img, nib.Gifti1Image):
        return img.agg_data()
    return img.get_fdata()
```

This feels generally useful and replaces duplicated load logic throughout the `analysis` module. I would pull this out, make it public and move it into the `images` module. The `what` argument is nice for debugging, but probably isn't needed.

This can then replace:
  - `nulls._resolve_data`
  - `nulls._load_one`
  - `analysis.images.load_gifti`

src/neuromaps_prime/analysis/parcellate.py
```diff
@@ -0,0 +1,191 @@
...
+import nibabel as nib
+import numpy as np
+from numpy.typing import ArrayLike
+
+__all__ = ["ParcellatedData", "parcellate"]
+
+
+Method = Literal["mean", "median", "sum", "std", "min", "max"] | Callable[
+    [np.ndarray], float
+]
```
kaitj:
```suggestion
ReduceMethod = Literal["mean", "median", "sum", "std", "min", "max"] | Callable[
    [np.ndarray], float
]
```

Would rename this to describe how this is used downstream (reducing values within a region); `Method` is a bit generic here.

src/neuromaps_prime/analysis/parcellate.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class ParcellatedData(NamedTuple):
```
kaitj:
```suggestion
class ParcelSummary(NamedTuple):
```

Would rename this too to reflect that it is a region-wise summary. Doesn't have to be this specific name, but just something to this effect.

src/neuromaps_prime/analysis/parcellate.py
```diff
@@ -0,0 +1,191 @@
...
+    if arr.ndim == 2:
+        return arr
+    if arr.ndim == 3:
+        return arr.reshape(-1, 1)
+    if arr.ndim == 4:
+        return arr.reshape(-1, arr.shape[-1])
+    raise ValueError(f"Unsupported {what} with {arr.ndim} dimensions.")
+
+
+def parcellate(
```
kaitj:
```suggestion
def parcel_reduce(
```

Similarly renaming reasoning as above here, but aligned to the suggested renamed output class.

src/neuromaps_prime/analysis/parcellate.py
```diff
@@ -0,0 +1,191 @@
...
+    elif callable(method):
+        reducer = method
+    else:
+        raise ValueError("method must be a string or a callable.")
+
+    data_arr = _load(data, what="data")
+    labels_arr = _load(parcellation, what="parcellation")
+
+    data_flat = _flatten(np.asarray(data_arr, dtype=float), what="data")
+    labels_flat = _flatten(np.asarray(labels_arr), what="parcellation")
```
kaitj:
```suggestion
    data_flat = _flatten(data_arr, what="data")
    labels_flat = _flatten(labels_arr, what="parcellation")
```

The `_load` call above already returns an `np.ndarray`

src/neuromaps_prime/analysis/parcels.py
```diff
@@ -0,0 +1,191 @@
...
+
+    data_flat = _flatten(np.asarray(data_arr, dtype=float), what="data")
+    labels_flat = _flatten(np.asarray(labels_arr), what="parcellation")
+
+    if labels_flat.shape[1] != 1:
+        raise ValueError(
+            "Parcellation must be a single volume or surface, got "
+            f"{labels_flat.shape[1]} features."
+        )
+    labels_1d = labels_flat[:, 0]
```
kaitj, line 143:
```suggestion
    labels_1d = np.rint(labels_flat[:, 0]).astype(np.int64)
```

Would just convert to int here and reference `labels_1d` throughout instead of having a separate `labels_int`, since 1d is only used to match shape.

tests/unit/analysis/test_parcellate.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def test_min_valid(self, surface_data):
+        data, labels = surface_data
+        data = data.copy()
+        data[[2, 3]] = np.nan  # region 1 keeps a single finite vertex
+        result = parcellate(data, labels, min_valid=2)
+        assert np.isnan(result.values[0])
+        assert np.isfinite(result.values[1:]).all()
+
+    @pytest.mark.parametrize(
```
kaitj: Missing a test here for `std`

kaitj (changes requested):
Largely looks good to me. Suggested some optimizations, as well as pulling out the loader and replacing duplicate code in other modules. There is some further optimization can be made with the `parcellate` function, but it wouldn't be a noticeable difference unless the data is huge. Also 1 missing test.

src/neuromaps_prime/analysis/parcels.py
```diff
@@ -66,6 +72,8 @@ def parcel_reduce(
     method: ReduceMethod = "mean",
     background: float | int | None = 0,
     min_valid: int = 1,
+    drop_nonfinite: bool = True,
```
kaitj, line 75:
```suggestion
    *,
    drop_nonfinite: bool = True,
```

To satisfy Ruff's FBT001 rule (no positional boolean args)

src/neuromaps_prime/analysis/parcels.py
```diff
@@ -138,32 +152,126 @@ def parcel_reduce(
...
+
+
+def _reduce_ndimage(
+    values: np.ndarray,
+    data_flat: np.ndarray,
+    labels_1d: np.ndarray,
+    unique: np.ndarray,
+    method: str,
+    min_valid: int,
+    drop_nonfinite: bool,
```
kaitj, line 208:
```suggestion
    *,
    drop_nonfinite: bool,
```

To satisfy Ruff, will also need to update the use of this to pass as a keyword-argument

src/neuromaps_prime/analysis/parcels.py
```diff
@@ -138,32 +152,126 @@ def parcel_reduce(
...
+
+
+def _reduce_callable(
+    values: np.ndarray,
+    data_flat: np.ndarray,
+    labels_1d: np.ndarray,
+    unique: np.ndarray,
+    method: Callable[..., float],
+    min_valid: int,
+    drop_nonfinite: bool,
```
kaitj, line 237:
```suggestion
    *,
    drop_nonfinite: bool,
```

to satisfy ruff

src/neuromaps_prime/analysis/parcels.py
```diff
@@ -138,32 +152,126 @@ def parcel_reduce(
...
+    sentinel = int(unique.max()) + 1
+    for f in range(data_flat.shape[1]):
+        col = data_flat[:, f]
+        finite = np.isfinite(col)
+        if drop_nonfinite:
+            eff_labels = np.where(finite, labels_1d, sentinel)
+        else:
+            eff_labels = labels_1d
+        counts = ndimage.sum_labels(finite.astype(float), labels_1d, index=unique)
+        summarised = np.asarray(reducer(col, eff_labels, unique), dtype=float)
```
kaitj:
```suggestion
        summarised = reducer(col, eff_labels, unique)
```

I don't think we need to use this `np.asarray` call here (might be wrong). I think the ndimage reducers already return as an array of type float.

kaitj (commented):
Great, looks good to me, just a few minor things to satisfy ruff and one redundancy. Once those are fixed, this looks good to merge in to me!

kaitj (approved):

PR #238, opened 2026-08-12
added notebook for aperture manuscript submission

tfunck:
## This PR:
Notebook that illustrates how to use many of the features in neuromaps-prime

Note: requires parcellation branch to be merged in before all the cells will run.


examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+    "\n",
+    "1. Load the template-space graph and see what maps are available\n",
+    "2. Transform an annotation from one space into another, including across species\n",
+    "3. Visualize the transformed map\n",
+    "4. Compare two maps with a statistically appropriate null model\n",
+    "5. Add your own template space and connect it to the graph\n",
+    "\n",
+    "**Setup.** You need Python and a container runtime (Docker, Podman, or Apptainer/Singularity).\n",
+    "You do **not** need Connectome Workbench or ANTs installed. Sections marked *offline* need neither\n",
+    "downloads nor a container.\n",
```
kaitj: This should be reframed as the tools "do not need to be strictly installed". With NiWrap, if the user already has the tool installed, it would just call that ("local"). Otherwise the container execution environments will handle the orchestration.

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "runner: docker (docker)\n"
+     ]
+    }
+   ],
+   "source": [
+    "# python3 -m pip install neuromaps-prime nilearn\n",
```
kaitj: I think there is probably value for having the pip install as its own cell that users can choose to skip in case installation is necessary. Can use the bash magic for this:

`!pip install neuromaps-prime nilearn`

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+    "\n",
+    "from pathlib import Path\n",
+    "import numpy as np\n",
+    "import nibabel as nib\n",
+    "\n",
+    "from neuromaps_prime.graph import NeuromapsGraph\n",
```
kaitj: I go back and forth for this as I prefer to have my libraries all imported at the very top of a notebook, but can also import as needed (which in this case I don't think we need these imports here.

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "NeuromapsGraph with 12 nodes and 42 edges\n",
+      "cache: /home/thomas-funck/.cache/neuromaps_prime\n"
+     ]
+    }
+   ],
+   "source": [
+    "graph = NeuromapsGraph(runner=\"auto\")\n",
```
kaitj: Just a note that "auto" is the default and doesn't need to be specified. You can also specify `verbose=1` to print out the metadata info after the transformer is called (in the cells below).

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "4d2694f1",
+   "metadata": {},
+   "source": [
+    "The labels are short type codes, optionally suffixed to distinguish variants:\n",
+    "\n",
+    "`CT` cortical thickness · `CV` curvature · `MM` myelin map · `SMM` smoothed myelin ·\n",
+    "`PC_*` parcellations · `RM_*` receptor maps · `SD` sulcal depth · `BM` brain mask\n",
```
kaitj: Could turn this into a markdown table

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+   "cell_type": "markdown",
+   "id": "6a46ebe9",
+   "metadata": {},
+   "source": [
+    "---\n",
+    "## 3. How transformations are found *(offline)*"
+   ]
+  },
+  {
+   "cell_type": "markdown",
```
kaitj, line 309: This cell probably would benefit from the graph figure highlighting these connections!

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+   "id": "1f5f97dc",
+   "metadata": {},
+   "source": [
+    "---\n",
+    "## 4. Transforming real data\n",
+    "\n",
+    "**[needs a container and downloads]**\n",
```
kaitj, line 391: "software' is probably more representative than "container" here.

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+    ")\n",
+    "\n",
+    "thickness_on_fsLR = graph.surface_to_surface_transformer(\n",
+    "    transformer_type=\"metric\",\n",
+    "    input_file=thickness_path,\n",
+    "    source_space=\"Yerkes19\",\n",
+    "    target_space=\"fsLR\",\n",
+    "    hemisphere=\"right\",\n",
+    "    source_density=\"32k\",\n",
+    "    target_density=\"32k\",\n",
```
kaitj, line 501: These density arguments are optional. Could be nice to see what happens if you don't specify these parameters.

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+    "    output_file_path=\"thickness_on_fsLR.func.gii\",\n",
+    ")\n",
+    "\n",
+    "print(myelin_on_fsLR)          # prints the real path\n",
+    "print(myelin_on_fsLR.metadata) # provenance for every hop\n",
```
kaitj: the `metadata` attribute actually holds an object that contains separate metadata for the nodes that are traversed, and the edges that it passes through, which is why it is printing out the object. If `verbose>=1`, this will automatically print (which includes any references and notes from the graph). You can also choose to save the metadata to disk!

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "542c4d8d",
+   "metadata": {},
+   "source": [
+    "---\n",
+    "## 5. Visualizing the result\n",
+    "\n",
+    "Neuromaps-PRIME writes standard GIFTI files and does not render surfaces itself, so any surface\n",
+    "plotting library will do. The example uses a custom matplotlib function, which is not a dependency and must be installed\n",
```
kaitj, line 569: `matplotlib` is a dependency (at least as of the latest dev version), though it probably doesn't need to be. Just a note, nothing to do be done.

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "nodes: 12 -> 13\n",
+      "reachable from MyLabTemplate: nothing yet\n"
+     ]
+    }
+   ],
+   "source": [
+    "my_space = {\n",
```
kaitj: There are a few ways we can do this. We can also use a combination of an instantiated graph and the models declared in `neuromaps_prime.graph.models`, though looking at this, that one seems more involved. The one benefit is you get validation as soon as you try to declare the model versus having to wait until registrationn to the graph. For example, just creating the node would look like:

```py
from neuromaps_prime.graph.models import Node

g = NeuromapsGraph()
my_lab_template = Node(
    name="MyLabTemplate",
    species="macaque",
    description="Example laboratory template, version 1.0",
    references=["Author A, et al., 2026. Journal Name. doi: 10.xxxx/xxxxx"],
    surfaces=[SurfaceAtlas(...), SurfaceAtlas(...), SurfaceAtlas(...), SurfaceAtlas(...)]
)
g.add_node(my_lab_template)
```

And something similar can be done with `g.add_edge(...)` to add an edge.

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+    "      graph.find_path(\"fsLR\", \"MyLabTemplate\", edge_type=\"surface_to_surface\") or \"NO ROUTE\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6d50dfa1",
+   "metadata": {},
+   "source": [
+    "**Provenance naming.** The key beneath `surfaces` names the method that produced the registration.\n",
+    "Several methods may connect the same pair of spaces, and naming them lets a user choose between them.\n",
```
kaitj, line 6987: Worth mentioning the `provider` argument here to the transformers for selection?

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+    "MyLabTemplate:\n",
+    "  surfaces:\n",
+    "    32k:\n",
+    "      annotation:\n",
+    "        MM:\n",
+    "          left:  /data/mylab/src-MyLabTemplate_den-32k_hemi-L_desc-MM_annot.func.gii\n",
+    "          right: /data/mylab/src-MyLabTemplate_den-32k_hemi-R_desc-MM_annot.func.gii\n",
+    "```\n",
+    "\n",
+    "Name files so they declare their own provenance:\n",
```
kaitj, line 7070: Is it worth noting that we suggest following a BIDS-like structure, but we don't enforce it? As long the user themselves can understand whats going on.

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+    "\n",
+    "```\n",
+    "src-{space}_den-{density}_hemi-{hemi}_desc-{code}_annot.{ext}\n",
+    "```\n",
+    "\n",
+    "To contribute back, fork the repository, add your YAML under `resources/`, confirm the graph loads\n",
```
kaitj, line 7077: One important thing to note is that if a path is provided, it is expected that a user can get access to it (e.g. its behind a DUA). We don't strictly enforce this, but graph will throw an error since the file cannot be found otherwise. Aside from that, urls are expected to be publicly available data that can be fetched from.

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,1499 @@
...
+      "text/plain": [
+       "<Figure size 640x480 with 1 Axes>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "from neuromaps_prime.analysis.parcellate import parcellate\n",
```
kaitj: Brief discussion offline about this, but wonder about using an existing library here to do the regional stats.

kaitj (commented):
Nice, its great to see the stats stuff in use, even for just a tutorial. There are a few things I flagged for wording and some suggestions and comments on use. Other than that looks great. Oh and I guess the ruff linter not being happy, but that is a an easy fix.

examples/neuromaps_prime_tutorial.ipynb
tfunck: I ended up putting all the standard library imports at the top, but I left the neuromaps_prime imports in the first cell where they're used. I think it it helps the user see what they need to do to import to get the code running

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,7196 @@
...
+    "import nibabel as nib\n",
+    "import matplotlib.pyplot as plt\n",
+    "import tempfile\n",
+    "\n",
+    "from neuromaps_prime.niwrap import resolve_runner\n",
+    "\n",
+    "from scipy.spatial import ConvexHull\n",
+    "\n",
+    "\n",
+    "runner_type, executable = resolve_runner(\"auto\")\n",
```
kaitj, line 132: I would maybe just have this call `resolve_runner()` -it defaults to "auto" and I assume this is just o show how the system would invoke tools. This can be explicitly set as part of the NeuromapsGraph arugments (e.g. `NeuromapsGraph(runner=...)`)

examples/neuromaps_prime_tutorial.ipynb
```diff
@@ -0,0 +1,7196 @@
...
+   "cell_type": "code",
+   "execution_count": 16,
+   "id": "973c3411",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "neuromaps_prime.analysis.surfaces.nulls.spins - Generating spin     0 of 1000\n",
```
kaitj, line 842: Now that this is verbose, it prints out which spin is being generated. For the purposes of the notebook, I wonder the results should be gathered and then printed outside of the loop? Otherwise the results can get lost in this long wall of text because of the verbosity.

kaitj (approved):
It looks good to me - just have 2 very minor comments, but nothing that would prevent the notebook from being merged at the moment.

